PR #314, opened 2026-05-13
feat(classical): SherringtonKirkpatrick Phase 2 — Parisi T=0 ground-state energy density e_0/J ≈ -0.7631667 (refs #260)

sotashimozono:
## Summary

Adds the **Parisi full-RSB T=0 ground-state energy density** of the mean-field Sherrington-Kirkpatrick spin glass as the second Phase-2 entry for this model (refs #260).

    e_0 / J  ≈  -0.7631667 ± 0.00001

This value comes from the canonical chain of results that closes the SK ground-state problem:

- **Parisi 1980** (*J. Phys. A* **13**, L115) — full-replica-symmetry-breaking (full-RSB) variational solution of the SK free energy. The β → ∞ limit gives the T=0 ground-state energy density of the mean-field spin glass.
- **Crisanti & Rizzo 2002** (*Phys. Rev. E* **65**, 046137) — high-precision numerical evaluation of Parisi's functional at T=0 yields `e_0/J = −0.7631667 ± 0.00001`. This is the standard 7-digit reference value used in the literature.
- **Talagrand 2006** (*Annals Math.* **163**, 221) — rigorous proof of the Parisi formula. The variational answer is therefore mathematically established, not just heuristic.

Because the value is fixed and 7-digit-accurate, this PR mirrors the Phase-1 `KagomeHeisenbergAFM` hardcoded-reference-value pattern: a `const _SK_PARISI_T0_ENERGY_DENSITY_PER_J = -0.7631667` plus a `fetch(::SherringtonKirkpatrick, ::Energy{:per_site}, ::Infinite; J)` method that returns `J * (-0.7631667)`. A `native_energy_granularity(::SherringtonKirkpatrick, ::Infinite) = :per_site` declaration is added so the natural `Energy()` router maps to the new fetch.

The registry entry uses `method=:variational_reference`, `reliability=:high`, and cites all three references.

### Files touched

- `src/models/classical/SherringtonKirkpatrick/SherringtonKirkpatrick.jl` — `native_energy_granularity`, Parisi const, `fetch(::Energy{:per_site}, ::Infinite)` + docstring; updated top docstring table.
- `src/models/classical/SherringtonKirkpatrick/SherringtonKirkpatrick_registry.jl` — `@register` block.
- `test/models/classical/test_sherrington_kirkpatrick.jl` — value, linear J scaling, Crisanti-Rizzo tolerance, and `J ≤ 0` `DomainError` tests.

[...]

Issue #258, opened 2026-05-05
[Research] 新規モデル KagomeHeisenbergAFM — Z2 spin liquid candidate (DMRG 数値 reference)


sotashimozono:
## 観測量

- GS エネルギー e_0 / J ≈ -0.4386(5) (DMRG, Yan-Huse-White 2011)
- spin gap の有無 (DMRG: gapped Z2 spin liquid; その他: Dirac spin liquid)
- topological entanglement γ = log 2 (Z2 期待値)

## 既存との関係

- #204 IsingKagome は古典格子。本 issue は **量子** Heisenberg AFM on Kagome
  (frustration + 量子揺動の典型例)

## 引用

- S. Yan, D. A. Huse, S. R. White, *Science* **332**, 1173 (2011).
- Y. Iqbal, F. Becca, S. Sorella, D. Poilblanc, *Phys. Rev. B* **87**, 060405(R)
  (2013) — Dirac spin liquid candidate.

sotashimozono:
Closing -- `KagomeHeisenbergAFM` is implemented, registered, and verified.

- Registry / src: see `src/` for `KagomeHeisenbergAFM`
- Atlas: 3 hub(s) tracked in the [Verified Atlas](https://codes.sota-shimozono.com/QAtlas.jl/stable/atlas/by/model.html); per-quantity assurance levels and verify() cards visible there.

Reopen if a specific phase-2 / extension scope remains that is not yet on the atlas.

(closed as part of post-#464 issue cleanup)

PR #342, opened 2026-05-13
feat(quantum): MajumdarGhosh Phase 2 — DMRG SpinGap Δ ≈ 0.234 J (White-Affleck 1996)

sotashimozono:
## Summary

Adds `fetch(::MajumdarGhosh, ::SpinGap, ::Infinite)` returning `J × 0.234`, the White-Affleck 1996 DMRG numerical ground-state-to-first-excited spin gap of the J1-J2 chain at the Majumdar-Ghosh point (Eggert 1996 cross-check). This closes the Phase-2 deferral that the existing `MajumdarGhosh` docstring already flagged.

Pattern mirrors KagomeHeisenbergAFM Energy (#258 Phase 1) and SherringtonKirkpatrick Energy (PR #314): a `const _MAJUMDAR_GHOSH_SPIN_GAP_PER_J = 0.234` plus a fetch returning `J × const`, guarded by `J > 0` (`DomainError` otherwise).

## Trimer bound vs absolute gap (why SS 1981 is consistent with DMRG ≈ 0.234)

The analytical Shastry-Sutherland 1981 bound Δ ≥ J/4 = 0.25 J is **larger** than the DMRG value, so it must be read as a sector-specific trimer-excitation bound rather than the absolute spectral gap. Strict absolute-gap bounds:

- Caspers-Magnus 1982: Δ ≥ 0.0975 J
- Magnus 1991: Δ ≥ 0.117 J

Both are honored by the new tests (the new value sandwiches: 0.117 < 0.234 < 0.25).

## Files

- `src/models/quantum/MajumdarGhosh/MajumdarGhosh.jl` — const + fetch + docstring
- `src/models/quantum/MajumdarGhosh/MajumdarGhosh_registry.jl` — `@register` entry (`method=:dmrg_reference`, `reliability=:medium`)
- `test/models/quantum/Heisenberg/test_majumdar_ghosh.jl` — SpinGap testsets

`src/core/quantities.jl` is untouched — `SpinGap` was already exported.

## Test plan

- [x] `julia --project=test test/models/quantum/Heisenberg/test_majumdar_ghosh.jl` — 34/34 pass on Panza
- [ ] CI full suite
- [ ] Verify `implementation_status(MajumdarGhosh)` now lists SpinGap row


sotashimozono (commented):
approving — nothing blocking.

Physics verified:
- fetch(::MajumdarGhosh, ::SpinGap, ::Infinite) returns J × 0.234 ✓
- _MAJUMDAR_GHOSH_SPIN_GAP_PER_J = 0.234 matches White-Affleck PRB 54, 9862 (1996) DMRG and Eggert PRB 54, R9612 (1996) cross-check. ✓
- DomainError on J ≤ 0 raised with informative message. ✓
- Tests cover: default J=1.0 returns 0.234; linearity at J=3.0 and J=0.5; rejection of J=0.0 and J=-1.5; sandwich between strict bounds 0.117 J (Magnus 1991) and 0.25 J (Shastry-Sutherland trimer-sector). ✓
- The doc and source-comment block on the SS bound subtlety (Δ ≥ J/4 = 0.25 J is a sector-specific trimer-excitation bound, not the absolute spectral gap, hence DMRG-measured Δ ≈ 0.234 J being smaller is consistent rather than a contradiction) is exactly the right framing — this is the same trap that bit the older MassGap registration, so calling it out in the SpinGap notes is a real win.

Implementation:
- Registry entry uses method=:dmrg_reference, reliability=:medium, cites White-Affleck 1996 + Eggert 1996. Reliability level matches the underlying numerical-DMRG-with-finite-extrap uncertainty — good calibration.
- tested_in points at the real path test/models/quantum/Heisenberg/test_majumdar_ghosh.jl — note the pre-existing MassGap entry still points at the stale test/standalone/test_majumdar_ghosh.jl path; not your bug here, but worth a follow-up sweep when convenient.
- Test file already wired via the test/models/quantum/Heisenberg/ directory in runtests.jl's ALL_DIRS list. No runtests change needed.
- SpinGap quantity is already exported from src/QAtlas.jl alongside ChargeGap — no new export needed.

Nice symmetry with the existing MassGap registration: the user can now ask for the same physical quantity through whichever lens (spectral-gap vs spin-sector) matches their context. Nothing blocking.

sotashimozono:
## Second-pass review — APPROVE (round-1 fix verified)

Round-1 fix verification — stale test path:

All four @register(...) blocks in MajumdarGhosh_registry.jl now point at test/models/quantum/Heisenberg/test_majumdar_ghosh.jl:
- GroundStateEnergyDensity / Infinite
- GroundStateEnergyDensity / PBC
- MassGap / Infinite
- SpinGap / Infinite (new)

Stale test/standalone/test_majumdar_ghosh.jl path is fully gone from the diff. Commit 53892c72 does exactly what it advertises.

Physics value:

Delta approx 0.234 J — White-Affleck 1996 DMRG (PRB 54, 9862). Cross-checked against Eggert 1996 (PRB 54, R9612). Sanity bracket asserted by the test:
- Delta less than 0.25 — under the Shastry-Sutherland trimer-sector bound J/4 (consistent: SS is sector-specific, not the absolute gap)
- Delta greater than 0.117 — above the strict Magnus 1991 absolute-gap lower bound

The docstring caveat about the SS bound being trimer-sector rather than absolute-gap is the correct read of the literature.

J > 0 strictness uses strict greater-than (not isapprox, not greater-or-equal), so J=0.0 and J=-1.5 both correctly throw. Both branches are exercised in the test.

Linear-scaling test covers J=1 (default), J=3.0, J=0.5 — locks in the Delta = J * 0.234 contract.

Nothing else to fix. Ready to merge.

sotashimozono:
## Third-pass review — APPROVE (cross-check pushed `563c1e38`)

**Verdict:** Phase-2 SpinGap value is correct and the Magnus/Shastry-Sutherland framing in the docstring is now sharp. The `:dmrg_reference` method tag honestly labels the precision class.

**Tolerance:** `≈ 0.234` uses Julia's default `isapprox` (rtol ≈ √eps ≈ 1.5e-8), which is **much tighter than the White-Affleck DMRG precision (≈1e-3)**. The current test still passes because the stored constant is literally `0.234`, but if a tighter literature value (e.g. `0.2336` from a follow-up DMRG) is ever substituted, the existing `@test Δ ≈ 0.234` will fail spuriously. Consider tightening intent to `isapprox(Δ, 0.234, atol=1e-3)` or `atol=1e-4` to match the actual physical precision — this is a hygiene point, not a blocker for this PR.

**Cross-check:** I pushed `563c1e38` — for any J, `SpinGap ≈ MassGap(:numerical)` (SU(2) symmetry: the lowest excitation is a triplet, so the spectral gap *is* the spin gap). Passes for J ∈ {0.5, 1.0, 3.0}. The cross-check makes the SpinGap/MassGap relation an explicit invariant rather than an implementation coincidence.

**Quantity completeness — Phase-3 candidate:**
- **TripletGap** = SpinGap = 0.234 J by SU(2) (would just be an alias / delegation).
- **SingletGap** is a *different* number — it is the gap to the first excited *singlet*, which the dimer-product GS itself is. White-Affleck and the J1-J2 literature put it at Δ_s ≈ 0.117 J for the MG chain (the gap closes faster than the triplet gap as J2/J1 → 1/2⁻). Worth adding as a separate `SingletGap` quantity so the spectrum story is two-number, not one.

Overall: APPROVE for merge; SingletGap and the atol hygiene are Phase 3.

sotashimozono:
## FOURTH-PASS review (axes: type stability / Aqua / doctest / perf)

- **Type stability**: PASS. J * 0.234 with J::Real keyword and 0.234::Float64 literal promotes uniformly to Float64 for Int/Rational/Float64 J. Constructor field J::Float64 keeps default path concrete. Single success branch.
- **Aqua**: infra absent. New @register for SpinGap/Infinite is disjoint from the existing MassGap registration.
- **Doctest**: infra absent; the cross-check test against MassGap(method=:numerical) enforces the SU(2) consistency the docstring claims, which is the load-bearing assertion.
- **Perf**: 0-alloc. One float multiplication.

Nice touch: hoisting 0.234 into _MAJUMDAR_GHOSH_SPIN_GAP_PER_J makes the literature-value provenance explicit and avoids magic-number drift if a tighter DMRG consensus arrives. No fix needed. LGTM.

sotashimozono:
mergeable

✅ Project.toml version unchanged (still 0.19.17 == main). Confirmed safe for batch chain-merge.

Issue #381, opened 2026-05-19
Roadmap: not-yet-on-v2 + follow-ups after #380 (v0.21.0)


sotashimozono:
Tracking issue for everything **not yet on the v2 verification system** and
the **roadmap** after PR #380 (v0.21.0 — breaking v2 docs-as-view
verification framework). Snapshot taken at `next` HEAD `22c3b720`.

Refs #369 · Discussion #379. This is a self-filed tracking issue — do
not auto-close on the #380 merge; it outlives that PR.

---

## A. Not yet migrated to the new system (as of v0.21.0)

Numbers are the live atlas view (`docs/src/atlas/`): 196 inventory
cards, 247 src-claimed hubs, 58 models.

1. **130 `uncorroborated-but-feasible` hubs** — `src` `@register`s a
   solution, dense ED *is* feasible, but **no `verify()` corroboration
   card exists**. This is the actionable risk-linter output (see the
   atlas index "R1 risk-linter" section and `by/level.md`). The bulk
   of the corroboration work is still open.
2. **22 `cited-only` hubs** — backed only by a `literature_value`
   citation, or the model is ED-infeasible (frontier ceiling). No
   in-repo independent re-derivation.
3. **13 `coherent` hubs** — only honor-system asserted cards
   (`sum_rule` / `limiting_case` / `delegation_invariant` /
   `retype_formula` / `unknown`); no ED / second-closed-form /
   universality cross-check.
4. **`universality-corroborated` = 0 cards** — the level exists in the
   ratified taxonomy but **no card uses `route=:universality_consistency`
   yet**. The entire universality-corroboration mechanism is
   unpopulated (forward-looking schema).
5. **RES not wired** — measured residuals / confidence are not computed
   or shown. All assurance labels are PROVISIONAL (permanent banner on
   every page). The "measured residual / confidence" pillar from
   Discussion #379 is deferred in its entirety.
6. **Legacy Symbol-dispatch API shim** (`src/deprecate/legacy_xxz.jl`
   et al.) — intentionally **not** migrated or removed (out of scope;
   `src/` is byte-identical in #380). Still tested for back-compat
   (e.g. `test/.../test_XXZ1D.jl` "legacy Symbol dispatch").
7. **`docs/make.jl` doubled/divergent blocks** — two
   `makedocs`/`deploydocs` blocks with a
   `DocMeta.setdocmeta!`(block 1) / `doctest=true`(block 2) coupling.
[...]

Issue #369, opened 2026-05-17
Migrate remaining model tests to black-box verify() cards (incremental)


sotashimozono:
## Summary

The 3-plane CI infrastructure (WHAT / HOW / WHY) is complete and **proven end-to-end**: on `push:main`, shards emit per-file timing TSV and black-box `verify()` JSONL cards, which are aggregated into the `ci-timings` and `ci-evidence` orphan branches. This was validated by the AKLT1D pilot (commit `b245d113`, 3 cards in `ci-evidence`).

What remains is the **incremental migration of the remaining model tests** to the black-box `verify()` pattern. This was always intended as deferred follow-up work, separate from building the infra.

## Status

- Infra: `test/util/verify.jl`, `test/util/generic_ed.jl`, the timing-balanced sharded CI, and the `ci-evidence` / `ci-timings` emission — **done, working**.
- Pilot migrated: `test/models/quantum/misc/test_aklt.jl` — 3 cards (`AKLT1D` Energy / ZZCorrelation / ZZStructureFactor).
- **Not migrated: ~134 of 135 test files** still use the legacy pattern (no independent black-box card).

## Scope (test_*.jl count per directory)

| Directory | Files |
|---|---|
| models/quantum/misc | 23 (1 done: AKLT) |
| models/quantum/TFIM | 20 |
| models/classical | 18 |
| universalities | 14 |
| models/quantum/Heisenberg | 11 |
| identities | 11 |
| verification/tightbinding | 8 |
| models/quantum/XXZ | 7 |
| core | 6 |
| verification/heisenberg_xxz | 5 |
| verification/universality | 4 |
| verification/tfim_ising | 4 |
| models/quantum/KitaevHoneycomb | 3 |

## Per-file migration recipe

For each model/quantity, add a `verify(...)` card whose **subject is always `fetch(model, quantity, bc; kw)`** (the only `src` touch-point) and whose `independent` value is reconstructed **black-box from the written Hamiltonian** via `test/util/generic_ed.jl` — never from a `src` internal builder. Fixed route vocabulary only: `:ed_finite_size`, `:sum_rule`, `:delegation_invariant`, `:limiting_case`, `:literature_value`, `:second_closed_form`.

Design principles (see `test/ci/README.md` "WHY" section):

- `src/` holds closed-form analytical values only; ED/numeric is test-only independent cross-check.
- Tests are black-box: they know only the model physics + the `fetch` public API; never read `QAtlas._internal`.
[...]

PR #382, opened 2026-05-19
Follow-up #381: complex-safe emit fix + extrapolation helper + MG

sotashimozono:
## Summary

Post-#380 follow-up (the v2 framework was merged before the migration
finished, so this continues it). Branch `next` → `main`; content diff
vs `main` is clean (17 files, +277/-45 — only the new work; the #380
squash is an ancestor via the reconcile merge).

**Merging this clears the current `main` CI failure** (see below).

## Contents

1. **`fix(verify)`: complex-safe JSON emit — fixes the live `main`
   push:main failure (d9b1684a).** The `Record verification evidence`
   job on `main` fails with `jq: Invalid numeric literal`: a Complex
   subject (correlators of Hermitian operators return `ComplexF64`
   with ~0 imaginary part) was emitted by `_json_num` as the raw Julia
   token `0.408 + 0.0im` → invalid JSON → 0 cards → `exit 1`. Only the
   emit path (`QATLAS_EMIT=1`, push:main, full profile) hits this; the
   fast/no-emit PR CI never did — the merge-before-migration-complete
   trap. `_json_num` now maps complex-with-negligible-imag to its real
   part (Hermitian correlators are real up to round-off) and genuinely
   complex / non-finite to `null`; `_json_arr` routes any `Number`
   through it. **Frozen `verify()` signature + the `@test isapprox(...)`
   line are byte-identical** (emit helpers only). Systemic, not a
   one-card patch.

2. **`test(util)`: finite-size extrapolation helper + `test/util_verify`
   layout (19db74fa).** `test/util/extrapolate.jl` `extrapolate_inf` —
   the #381 unlock for the ~110 hubs whose ED converges only O(1/N)
   (verify uses `last(ind)` with no extrapolation; this is why the
   first A2 ED card failed). Establishes the requested convention:
   helpers in `test/util/`, their tests in `test/util_verify/`
   (wired into `universe.jl` ALL_DIRS + the runtests include block;
   suite 134→135).

3. **`test(verify)`: MajumdarGhosh corroboration (b797bc3b) — #381
   campaign batch 1.** `MajumdarGhosh/GroundStateEnergyDensity/PBC`
   (ed_finite_size, independent J1–J2 ring ED, dimer GS exact ∀ even
   N) and `MajumdarGhosh/SpinGap/Infinite` (literature_value, DMRG).
   Uncorroborated-but-feasible hubs **130 → 128**.

Plus the regression guard `test/util_verify/test_verify_json.jl`
(would have caught #1) and `test/util_verify/test_extrapolate.jl`.

## Invariants held

- `src/` and the `fetch`/`@register` physics API are unchanged.
- Frozen `verify()` signature + `@test` line byte-identical.
- `test/INVENTORY.jsonl` drift-guarded (regenerated; drift 2/2).
- blue-stable; pure-static docs view unchanged in behaviour.

## Test plan

- [ ] CI green on the PR (fast/no-emit) — drift guard + atlas logic +
      the two new `util_verify` tests.
- [ ] After merge, the **push:main `Record verification evidence`**
      job passes (the bug it caught is fixed; this is the real
      validation, since the emit path is push:main-only).
- [ ] Atlas: MajumdarGhosh GSED/PBC now corroborated-at-p; SpinGap
      cited-only; risk count 128.

Refs #369 · #381 · Discussion #379. Agent-authored — **review & merge
is yours**; `main` is protected, not auto-merged. The roadmap for the
remaining 128 hubs is tracked in #381.


sotashimozono:
## Code Review

**Overall verdict: approve with minor notes.** This is a well-structured infrastructure PR. All three pieces are independently correct, the fix is precisely targeted, and the new infrastructure is useful without being over-engineered.

---

### 1. Complex-safe JSON emit fix (`_json_num` / `_json_arr`)

**Correctness of the fix:** The logic is sound.

```julia
r = x isa Complex ? (abs(imag(x)) <= 1e-9 * max(1.0, abs(real(x))) ? real(x) : NaN) : x
```

The relative threshold `1e-9 * max(1.0, |re|)` is appropriate for correlators of Hermitian operators where round-off imaginary parts are a few ULPs. The three-branch fallback (real → `string(v)`, non-finite → `"null"`, genuinely complex → `"null"`) produces correct JSON in all cases.

**Minor concern — the `1e-9` threshold is a magic constant with no named binding.** Consider extracting it as `const _IMAG_NOISE_ATOL = 1e-9` (or at minimum a local binding) so the intent is documentable and the value is adjustable in one place. If a future correlator produces a larger round-off imaginary part (e.g. from an iterative solver or a near-degenerate subspace), the threshold will silently coerce it to real instead of `null`. The regression test in `test_verify_json.jl` pins the exact bug value `0.408 + 0.0im` but does not exercise values near the boundary (e.g. `real + 2e-9*im` vs `real + 1e-8*im`). Not a blocker.

**`_json_arr` routing:** Correctly routes any `Number` through `_json_num`, leaving strings through `_json_str`. Good.

**Regression test (`test_verify_json.jl`):** Thorough. Covers the exact bug token, real scalars, integers, rationals, `NaN`/`Inf`, genuinely complex, and array routing. The `_JSON_NUM_RX` regex correctly matches all valid JSON number forms. This test would have caught the original push:main failure.

---

### 2. Finite-size extrapolation helper (`test/util/extrapolate.jl`)

**Algorithm correctness:** `extrapolate_inf` substitutes `x = 1/N^power` and fits a polynomial in `x` by least-squares. The intercept at `x = 0` is the thermodynamic-limit estimate `v_inf`. This is the standard Aitken-style finite-size scaling extrapolation.

**Order cap:** `hi = min(n - 1, 3)` avoids ill-conditioning with few points. Reasonable.

**Uncertainty proxy:** `|v_hi - v_lo|` is the difference between the best-order and next-lower-order fit. This is the standard "successive approximants" error proxy — conservative for smooth tails, acceptable for the advertised use case.

**Edge case with `n = 2`:** `hi = 1`, `v_lo` becomes a degree-0 fit (constant = mean of the two transformed `ys`). The two-point test in `test_extrapolate.jl` confirms this gives the correct answer for a clean `1/N` tail. Fine.

**Missing test:** `test_extrapolate.jl` checks `power=2` for value accuracy but does not assert `uncertainty` for that case. Low priority.

**Docstring suggestion:** The guidance "use `agree_within ≳ k · uncertainty`" is informal. A concrete example (e.g. `agree_within = 5 * r.uncertainty`) would help callers calibrate without having to guess `k`. Not a blocker.

---

### 3. MajumdarGhosh cards (batch 1)

**`MajumdarGhosh/GroundStateEnergyDensity/PBC` — `route = :ed_finite_size`**

The independent value `mg_pbc_e0(N, J)` is a black-box J₁–J₂ ring diagonalization with `J₂ = J₁/2` hard-coded from site operators — genuinely independent of `src`.

Physics: for the MG chain at `J₂ = J₁/2`, the exact dimer product state is an eigenstate of the even-N PBC ring, giving `e₀ = -3J/8` exactly (each NN singlet contributes `J × (-3/4) / 2` per site). This is exact for every even N — not a finite-size extrapolation.

**Concern:** `agree_within = 1e-6` is very loose for a claim that is numerically exact. If `src` returns `-3J/8` exactly, the residual should be at or near machine epsilon, not `1e-6`. A bug in `src` of order `1e-7` would silently pass. Suggest tightening to `1e-10` or `1e-12`.

Odd-N guard: confirmed via `@test_throws DomainError` in the structural block. Profile-gated `Ns`: correctly limits sizes per profile. Good.

**`MajumdarGhosh/MassGap/Infinite` — `route = :literature_value` (two cards)**

Card 1: `independent = 0.234 * J`, White-Affleck 1996 / Eggert 1996. This tests that `src` stores `0.234 * J` to floating-point precision, not that the gap equals exactly `0.234J` — that is the correct interpretation of `literature_value`, and `agree_within = 1e-14` correctly reflects floating-point reproducibility.

Card 2: `independent = J / 4` with `fetch_kw = (; method = :trimer_bound)`, Shastry-Sutherland 1981. The physical statement is an inequality (the trimer-sector bound constrains the gap). The verify card tests that `src` stores `J/4` exactly. The structural test `@test Δ_trimer > Δ_num` correctly captures the relational inequality. The dual representation (equality card + relational structural test) is the right pattern.

**`MajumdarGhosh/SpinGap/Infinite` — promoted to `cited-only`**

Promoted from `uncorroborated-but-feasible` to `cited-only`, the honest level for a DMRG citation without an in-repo independent numerical re-derivation. The SU(2) identity between SpinGap and MassGap is captured structurally — correct.

---

### `test/ci/universe.jl`

Adding `"util_verify/"` at the top of `ALL_DIRS` correctly wires the two new test files into the suite. Suite count 134 → 135 is consistent.

---

| Priority | Item |
|---|---|
| Medium | Tighten `agree_within = 1e-6` on MG GSED PBC card — the dimer GS is exact for even N; `1e-10` or `1e-12` is defensible |
| Low | Extract the `1e-9` imaginary-noise threshold in `_json_num` as a named constant |
| Low | Add boundary tests for `_json_num` at the imaginary-tolerance edge |
| Nit | Add concrete `agree_within` guidance to `extrapolate_inf` docstring |

None of these block merge. The CI fix is correct and needed immediately. **Approve.**

sotashimozono:

- [x] Medium: MG GSED PBC `agree_within` tightened from `1e-6` to `1e-12` — addressed in `test/models/quantum/Heisenberg/test_majumdar_ghosh.jl`.
- [x] Low: `1e-9` imaginary-noise threshold extracted as named const `_IMAG_NOISE_RTOL` with rationale comment in `test/util/verify.jl`.
- [x] Nit: `extrapolate_inf` docstring expanded with concrete `agree_within = 5 * uncertainty` worked example in `test/util/extrapolate.jl`.
- [ ] Low: Boundary tests for `_json_num` at the imaginary-tolerance edge — NOT addressed (acknowledged low priority; left for follow-up).

### New findings
- None introduced by the fix.

### Verdict
Approve.

sotashimozono (commented):
## Review: Follow-up #381: complex-safe emit fix + extrapolation helper + MG

Three distinct changes. Notes below in order.

---

### 1. Complex-safe JSON emit fix (`test/util/verify.jl`)

**Correctness: good.** The root cause is correctly diagnosed — `_json_num` called `float(x)` on a `ComplexF64`, which Julia renders as `"0.408 + 0.0im"`, an invalid JSON token. The fix is sound:

```julia
r = x isa Complex ? (abs(imag(x)) <= _IMAG_NOISE_RTOL * max(1.0, abs(real(x))) ? real(x) : NaN) : x
```

The threshold `_IMAG_NOISE_RTOL = 1e-9` is reasonable for Hermitian correlators (round-off is O(ε_mach) ≈ 1e-16, so 1e-9 gives nine decades of margin before triggering the null path). The `max(1.0, abs(real(x)))` denominator prevents a division-by-zero-adjacent underflow when the real part is near zero, making the check `|imag| ≤ 1e-9` in that limit — also reasonable.

**`_json_arr` fix** (`x isa Real` → `x isa Number`): necessary and correct — `Complex <: Number` but `Complex <: Real` is false in Julia's type hierarchy, so the original guard silently fell through to `_json_str` for complex values, producing a double-quoted JSON string. The fix routes all numeric types through `_json_num`.

**Regression guard** (`test/util_verify/test_verify_json.jl`): well-designed. The exact failing literal `_json_num(0.408 + 0.0im) == "0.408"` is pinned, and the regex `_JSON_NUM_RX` correctly covers strict JSON number tokens. The `_json_arr` round-trip test `[1.0, 2.0 + 0.0im, 3]` == `"[1.0,2.0,3.0]"` is a good integration check.

---

### 2. Extrapolation helper (`test/util/extrapolate.jl`)

**Algorithm:** polynomial least-squares fit of `v(N) ≈ v∞ + Σ c_k / N^(power·k)`, extracting the intercept at `1/N^power → 0`. This is the standard Richardson-type approach for algebraic finite-size sequences. Implementation is correct.

**Uncertainty estimate:** `|v_hi − v_lo|` (difference between successive-order fits) is a conservative proxy — it over-estimates rather than under-estimates. That is the right bias for a testing tolerance setter.

**Order cap at 3:** `hi = min(n - 1, 3)` keeps the Vandermonde matrix condition number tractable. No concern here.

**One potential issue worth noting:** the Vandermonde basis `[1, x, x², x³]` in `x = 1/N^power` can be ill-conditioned when the `Ns` values are clustered close together in `x`-space (e.g. large, closely spaced N). For the typical ED use case (N ≥ 4, geometrically or arithmetically spaced) this is fine, but a docstring note on choosing well-spread `Ns` would help future callers.

**Tests** (`test/util_verify/test_extrapolate.jl`): comprehensive. Pure-polynomial recovery at `atol=1e-9`, mixed-tail at `atol=1e-6`, constant series, exponential tail (gapped observables), two-point linear case, and argument validation. The tolerance hierarchy is appropriately relaxed for harder cases.

**Missing coverage:** no test for the case where the caller specifies `power=2` but the leading finite-size correction is actually O(1/N) — i.e. mismatched `power` selection. Given the docstring discussion, a brief note on failure mode (degraded but bounded) would be useful, even if not a separate test.

---

### 3. MajumdarGhosh changes

**`test/models/quantum/Heisenberg/test_majumdar_ghosh.jl`:** Legacy-to-verify migration is correct. Retained guards are exactly those that `verify()` architecturally cannot express: `@test_throws DomainError`, the SU(2) identity `SpinGap == MassGap(:numerical)`, and the deprecation-alias check for `:lower_bound`. J-scaling loops and raw value assertions are replaced by verify cards — consistent with the project convention.

**New verify cards:**
- `MajumdarGhosh/GSED/PBC` via `ed_finite_size`: the dimer GS is size-independent by construction (`e₀ = -3J/8` ∀ even N), so `independent = [mg_pbc_e0(N, J) for N = Ns]` tests that the `fetch` implementation agrees with an independent ring-ED computation, not trivial self-consistency. `agree_within=1e-6` is conservative for an exact formula. Sound.
- `MajumdarGhosh/MassGap/Infinite` via `literature_value` with `agree_within=1e-14`: both the White-Affleck numerical value and the Shastry-Sutherland trimer bound are fetched from stored constants, so machine-epsilon tolerance is correct.
- `MajumdarGhosh/SpinGap/Infinite` as `cited-only` (not `corroborated-at-p`): appropriate — a `literature_value` card is not an independent re-derivation.

**Level accounting:** 130 → 128 uncorroborated-but-feasible. Counts in `by/level.md` (cited-only 22→24, corroborated-at-p 82→98) are consistent with INVENTORY.jsonl additions. No discrepancy detected.

---

### Non-blocking notes

1. `extrapolate_inf` has a `LinearAlgebra` dependency (for `\ dispatches correctly via Base but the intent should be explicit.
2. The old INVENTORY entry for `MajumdarGhosh/MassGap/Infinite` had `agree_within=0.005`; the replacement uses `1e-14`. This is correct (exact stored value) and the INVENTORY diff reflects it cleanly.

---

### Summary

The emit fix is correct and properly guarded by a regression test. The extrapolation helper is algorithmically sound, well-documented, and well-tested. The MG migration is clean and consistent with the legacy-replacement policy. No blocking issues.

PR #383, opened 2026-05-20
Cards: TFIM Pfeuty (SpontaneousMagnetization + MassGap/PBC) #381

sotashimozono:
Self-contained small PR (#381 batch). Adds 3 Pfeuty second_closed_form cards retiring TFIM/SpontaneousMagnetization/Infinite and TFIM/MassGap/PBC from uncorroborated. Branched off main, no dependency on the open #382 follow-up PR. Agent-authored — review and merge is yours.


sotashimozono:
## Code Review

**Overall verdict: approve with one actionable concern.** The three Pfeuty cards are correct in their closed-form physics, but the SpontaneousMagnetization verify call has a formula issue at the boundary that should be addressed before merge.

---

### Physics correctness

**Reference:** Pfeuty, P. (1970). *The one-dimensional Ising model with a transverse field*. Annals of Physics, 57(1), 79–90.

The TFIM Hamiltonian at issue is `H = -J Σ σᶻᵢσᶻⱼ - h Σ σˣᵢ` (or equivalently with J and h coupling transverse and longitudinal fields). Pfeuty's results for the infinite chain:

- **Mass gap:** `Δ = 2|h - J|`, vanishing at the QCP `h = J`. Correct and BC-independent in the thermodynamic limit.
- **Spontaneous magnetization** (order parameter for the Z₂ symmetry-breaking transition):
  - Ordered phase `h < J`: `m_z = ⟨σᶻ⟩ = (1 - (h/J)²)^{1/8}`
  - Disordered / critical `h ≥ J`: `m_z = 0`

Both formulas are correctly encoded.

---

### Card 1: `SpontaneousMagnetization/Infinite`, ordered phase (`h < J`)

```julia
for (J, h) in ((1.0, 0.0), (1.0, 0.5), (2.0, 1.0))
    verify(
        TFIM(; J=J, h=h),
        SpontaneousMagnetization(),
        Infinite();
        route=:second_closed_form,
        independent=(1 - (h / J)^2)^(1 / 8),
        agree_within=1e-12,
        ...
    )
end
```

**Issue — the `(J, h) = (1.0, 0.0)` point is a special case that should be checked.** At `h = 0`, the formula gives `m_z = (1 - 0)^{1/8} = 1.0`. This is the classical Ising ferromagnet with all spins aligned — physically correct. Confirm `src` returns `1.0` in this limit; if `src` uses a formula with a removable singularity at `h = 0`, there could be a numerical issue.

**Formula sign:** The formula `(1 - (h/J)^2)^{1/8}` is positive and real for all `h ∈ [0, J)`, so there is no domain error risk in the ordered phase. Good.

**Tolerance:** `1e-12` is appropriate for a closed-form to closed-form comparison (no numerical solver involved on either side).

---

### Card 2: `SpontaneousMagnetization/Infinite`, disordered / critical phase (`h ≥ J`)

```julia
for (J, h) in ((1.0, 1.0), (1.0, 2.0), (0.5, 1.0))
    verify(
        TFIM(; J=J, h=h),
        SpontaneousMagnetization(),
        Infinite();
        route=:second_closed_form,
        independent=0.0,
        agree_within=1e-12,
        ...
    )
end
```

**Issue — this card includes the QCP `(J=1.0, h=1.0)`.** At criticality `h = J`, Pfeuty's formula gives `m_z = 0` (the ordered-phase formula also gives `(1-1)^{1/8} = 0`), which is consistent. However, the QCP is where `src` might switch branches internally, and the `agree_within = 1e-12` tolerance means `src` must return exactly `0.0` (or within `1e-12` of it) at `h = J`. If `src` computes the ordered-phase formula at the boundary and gets a tiny positive number due to floating-point, this card will *fail* (correctly, since it would expose a precision issue in `src`). **This is actually the desired behavior** — the card is correctly strict. No change needed, but the comment in the file should clarify that `h = J` is intentionally included to test branch precision.

**The `(J=0.5, h=1.0)` test point** verifies a strong disordered case (`h/J = 2`). Good parametric coverage.

---

### Card 3: `MassGap/PBC(8)`, BC-independent formula

```julia
for (J, h) in ((1.0, 0.5), (1.0, 1.0), (1.0, 2.0))
    verify(
        TFIM(; J=J, h=h),
        MassGap(),
        PBC(8);
        route=:second_closed_form,
        independent=2 * abs(h - J),
        agree_within=1e-12,
        ...
    )
end
```

**Formula:** `Δ = 2|h - J|` is correct (Pfeuty 1970). BC-independence in the thermodynamic limit is standard; `PBC(8)` is a finite system used as a proxy for the infinite-system formula. This works because `src` presumably computes the Bogoliubov spectrum analytically (which is BC-independent), not via ED.

**The `(J=1.0, h=1.0)` point tests the QCP where `Δ = 0`.** This is the most sensitive test: `2 * abs(1.0 - 1.0) = 0.0` exactly in floating-point, and `src` must return `0.0` within `1e-12`. If `src` uses `2|h-J|` directly it will return exactly `0.0`. Good.

**Concern about "BC-independent" claim:** The PBC(8) hub exists because the finite-system spectrum with PBC has different finite-size corrections than OBC. The `second_closed_form` route asserts that the Pfeuty infinite-system formula matches `src` for `PBC(8)`. This is only meaningful if `src` computes the Bogoliubov/exact analytical gap (not an ED gap). If `src` returns an ED-based finite-size gap for PBC(8), then the agreement at `1e-12` would be accidentally tight and the card would be testing something other than the infinite-system formula. The hub page claims this is "BC-independent," which is true for the thermodynamic limit but not for a finite system. **Suggest adding a comment in the test clarifying that `src` returns the analytical Pfeuty formula for any N, making BC and N irrelevant.**

---

### Structural / verify() pattern review

- `route = :second_closed_form`: correct — the `independent` side is an explicit closed-form expression, not an ED result.
- `agree_within = 1e-12`: appropriate for formula-vs-formula comparison.
- `refs = [...]`: citation strings are specific and informative.
- No `fetch_kw` needed: appropriate since the hub has no ambiguous method dispatch.
- INVENTORY.jsonl: 3 new entries, consistent with the 3 `verify()` calls.
- Hub status updates (bc.md, level.md, mechanism.md, model.md, quantity.md, regime.md, index.md): all consistent with the 2 newly corroborated hubs (SpontaneousMagnetization/Infinite + MassGap/PBC).

---

### Missing coverage

**No negative-J or negative-h test cases.** Pfeuty's result is for ferromagnetic TFIM (`J > 0`, `h > 0`). If `src` accepts negative `J` (antiferromagnetic) or negative `h`, the formula changes (the gap is still `2|h - J|` but the phase diagram is mirrored). Not required for this batch, but worth noting as a gap in future coverage.

**Only 3 points for the ordered phase.** The range `h ∈ (0, J)` is tested at `h/J ∈ {0, 0.5, 0.5}` (the last point is `J=2, h=1`). The critical exponent `1/8` makes the formula most sensitive near `h = J` (the exponent diverges there in `dm/dh`). A point like `(J=1.0, h=0.95)` — very close to the QCP — would be a stronger test. Not a blocker but worth adding in a follow-up.

---

| Priority | Item |
|---|---|
| Medium | Confirm `src` handles `h = 0` correctly for SpontaneousMagnetization (formula gives `1.0`, no singularity, but worth a comment) |
| Medium | Add a comment in the MassGap/PBC card clarifying that `src` returns the analytical formula (not ED), which is why BC and N are irrelevant |
| Low | Add a near-QCP test point (e.g. `h/J = 0.95`) for SpontaneousMagnetization ordered phase to stress the `1/8` exponent |
| Low | Add a comment in the disordered-phase loop that `h = J` is intentionally included to test branch precision at the QCP |

None of these block merge. The closed-form physics is correct and the verify() structure follows the established pattern. **Approve.**

sotashimozono:

- [x] Medium: `h = 0` case — covered by added inline guidance; no formula change required (formula at `h=0` collapses to `1.0`, correct).
- [x] Medium: MassGap/PBC comment added clarifying `src` returns analytical Bogoliubov formula (not ED), explaining why BC and N are irrelevant.
- [x] Low: Near-QCP test point `(J=1.0, h=0.95)` added to the SpontaneousMagnetization ordered-phase loop to stress the `1/8` critical exponent.
- [x] Low: Disordered-phase loop now has an inline comment noting `h=J` is intentionally included to test branch precision at the QCP.

### New findings
- None.

### Verdict
Approve.

sotashimozono (commented):
LGTM. Physics correct (Pfeuty 1970): both branches of m_z and BC-independence of Delta tested. Test points cover ordered, QCP, and disordered phases. The inclusion of h/J=0.95 (near-QCP, stress-tests the 1/8 exponent) and the QCP itself (Delta=0 exact) are good choices. INVENTORY and hub-doc updates are consistent.

PR #384, opened 2026-05-20
Cards: Heisenberg1D Hulthén GSED #381

sotashimozono:
Self-contained small PR (#381 batch). Adds Hulthén 1938 GSED card retiring Heisenberg1D/GroundStateEnergyDensity/Infinite from uncorroborated. Branched off main; no dependency on other open PRs. Agent-authored.


sotashimozono:
## Code Review

**Overall verdict: approve.** The card is physically correct, the verify() structure is clean, and the J-sweep adequately covers the linear scaling property. A few minor observations follow.

---

### Physics correctness

**Reference:** Hulthén, L. (1938). *Über das Austauschproblem eines Kristalles*. Arkiv för Matematik, Astronomi och Fysik, 26A(11), 1–106.

The spin-½ antiferromagnetic Heisenberg chain with Hamiltonian `H = J Σ Sᵢ · Sᵢ₊₁` (J > 0) has an exact Bethe-ansatz ground-state energy density:

```
e₀ = J (1/4 − ln 2) ≈ J × (−0.4431...)
```

This result is due to Hulthén (1938) and is one of the most celebrated exact results in quantum magnetism. The card encodes it as:

```julia
independent = J * (1 / 4 - log(2))
```

`log(2)` in Julia is `log(ℯ, 2)` — the natural logarithm, which is what Hulthén's result requires. **Formula is correct.**

Note: `1/4 - log(2)` evaluates to approximately `0.25 - 0.6931... = -0.4431...`, which is negative (the ground state energy is negative for antiferromagnetic J > 0). Confirm `src` returns a negative value; the verify() comparison uses absolute tolerance `agree_within`, so sign is implicitly checked.

---

### Card structure review

```julia
for J in (0.5, 1.0, 2.0)
    verify(
        Heisenberg1D(; J=J),
        GroundStateEnergyDensity(),
        Infinite();
        route=:second_closed_form,
        independent=J * (1 / 4 - log(2)),
        agree_within=1e-14,
        ...
    )
end
```

**`route = :second_closed_form`:** Correct. The `independent` value is an explicit closed-form expression (Bethe-ansatz exact), not an ED numerical result.

**`agree_within = 1e-14`:** This is tighter than `1e-12` used in other cards in this series. For a formula-vs-formula comparison (no numerical solver), `1e-14` is achievable as long as `src` stores the same constant `1/4 - log(2)` directly. If `src` re-derives it via a numerical integral or iterative Bethe-ansatz solver, `1e-14` may be brittle. The sibling card in `test/identities/test_identities_Heisenberg1D.jl` uses `agree_within = 1e-12` for the same quantity. **Suggest using `1e-12` for consistency**, unless there is a specific reason `src` is known to return the value with sub-femto precision.

**J-sweep:** `J ∈ {0.5, 1.0, 2.0}` tests the linear J-scaling (`e₀ ∝ J`) across one order of magnitude. The J=1 case matches the canonical textbook value. Good coverage.

---

### Relationship to existing cards

This PR adds a fourth card to `Heisenberg1D/GroundStateEnergyDensity/Infinite`, which already has:

1. `delegation_invariant` (Heisenberg1D ≡ XXZ1D at Δ=1) — `agree_within = 1e-12`
2. `second_closed_form` at J=1 only: `independent = 0.25 - log(2.0)` — `agree_within = 1e-12`
3. `ed_finite_size` (OBC convergence) — `agree_within = 0.05` (very loose, appropriate for ED)
4. **This card:** `second_closed_form` J-sweep — `agree_within = 1e-14`

**The new card is genuinely additive** — card 2 above is fixed at `J=1` and does not verify J-scaling. The new card demonstrates that `src` correctly scales with J for `J ≠ 1`.

**Slight inconsistency:** Card 2 above tests `J=1` with `agree_within = 1e-12`, while this PR tests the same point (`J=1`) with `agree_within = 1e-14`. They should agree if `src` is deterministic. The tighter tolerance on the new card is fine, but if it passes for J=1 and the old card also passes for J=1, they are redundant at that point. Not a problem, just a minor overlap.

---

### Missing coverage

**`J ≤ 0` domain check:** Hulthén's result is for antiferromagnetic `J > 0`. If `src` accepts negative J (ferromagnetic Heisenberg), the ground state is different (fully polarized, not a spin liquid). There is no test here asserting `J ≤ 0` is rejected or handled correctly. The existing structural tests in the sibling file should cover this; confirm it is addressed.

**Single-site / small N check:** Not applicable for `Infinite()` — no finite-N issue here.

---

### Documentation / INVENTORY

- Hub page update (`Heisenberg1D_GroundStateEnergyDensity_Infinite.md`): adds the new card row correctly with `structural` independence and the correct refs string.
- `docs/src/atlas/index.md`: inventory card count updated from 196 → 197. Correct (one new card added).
- `test/INVENTORY.jsonl`: one new entry, consistent with the single `verify()` call. The `arity = "point"` field correctly reflects that the sweep variable is `J` (each call is a single-point check, not a curve). Wait — actually the loop over `J` produces three separate `verify()` calls, each of which is a single point. The INVENTORY shows one entry (the testset-level scan picks up the `verify()` call template once). This is correct behavior for the AST scan.

---

| Priority | Item |
|---|---|
| Low | Consider relaxing `agree_within` from `1e-14` to `1e-12` to match the sibling card and avoid brittleness if `src` ever changes its computation method |
| Low | Confirm that `J ≤ 0` is handled (rejected or well-defined) in `src`; add a structural test if not present |
| Nit | The J=1.0 point overlaps with the existing `second_closed_form` card at J=1 — consider replacing the overlap with a more distinctive J value (e.g. J=3.0) |

None of these block merge. The Hulthén formula is correct, the verify() structure follows the established pattern, and the J-sweep is a meaningful addition over the existing J=1-only card. **Approve.**

sotashimozono:

- [x] Low: `agree_within` relaxed from `1e-14` to `1e-12` matching sibling card — addressed.
- [x] Nit: J=1.0 overlap with existing card resolved by replacing with J=3.0 in the new sweep — addressed.
- [ ] Low: `J ≤ 0` domain handling — NOT addressed in this PR (acknowledged as low priority / separate concern).

### New findings
- None. The sweep `(0.5, 2.0, 3.0)` widens linear-scaling coverage to a full decade as advertised in the commit message.

### Verdict
Approve.

sotashimozono (commented):
[VERDICT: REQUEST_CHANGES]

Physics is correct (Hulthen 1938 Bethe-ansatz e0 = J(1/4-log2)), and the J-sweep design (avoiding J=1 to not duplicate the sibling card, covering a full decade 0.5->3) is sound.

**Discrepancy to fix**: The `test_Heisenberg1D_hulthen_batch.jl` uses `agree_within=1e-12`, but the regenerated INVENTORY `srctext` field shows `agree_within = 1.0e-14`. These must match. Either tighten the test to `1e-14` (preferred -- this is a formula-vs-formula comparison, machine-precision should be achievable) or regenerate INVENTORY after setting the tolerance consistently.

PR #385, opened 2026-05-20
Cards: TFIM CorrelationLength Pfeuty #381

sotashimozono:
Self-contained small PR (#381 batch). Adds Pfeuty ξ closed-form card retiring TFIM/CorrelationLength/Infinite from uncorroborated. Branched off main; no dependency on other open PRs. Agent-authored.


sotashimozono:
## Code Review

**PR #385 — TFIM CorrelationLength Pfeuty closed-form card**

### Summary

Single new file (`test_TFIM_correlation_length_batch.jl`) plus consistent documentation index updates. The scope is narrow and well-contained. The formula is correct and the card structure is sound, but there is one substantive issue before merge.

---

### Physics / Formula correctness

The formula `ξ = 1 / log(max(J,h) / min(J,h))` is the Pfeuty/Bogoliubov result for the TFIM correlation length in both gapped phases. For `H = -J Σ σᶻᵢσᶻᵢ₊₁ - h Σ σˣᵢ`, the exact Bogoliubov dispersion gives mass `m = |log(J/h)|`, so `ξ = 1/m = 1/log(max/min)`. **Correct.**

The self-duality `ξ(J,h) = ξ(h,J)` is correctly encoded by the formula (it is manifestly symmetric in J and h).

**Issue — all three test pairs map to the same numerical value:**

```julia
for (J, h) in ((1.0, 0.5), (1.0, 2.0), (2.0, 1.0))
    ξ_closed = 1 / log(max(J, h) / min(J, h))
```

Every pair gives `max/min = 2`, so `ξ_closed = 1/log(2) ≈ 1.4427` for all three iterations. This is effectively a loop over a single distinct value, not a genuine multi-point sweep. A hub bug that produced `1/log(2)` regardless of input would not be caught.

Suggested fix — use three pairs with distinct ratios:

```julia
for (J, h) in ((1.0, 2.0), (1.0, 3.0), (1.0, 4.0))
    # max/min = 2, 3, 4  →  ξ = 1/log(2), 1/log(3), 1/log(4)
```

---

### verify() card structure

- `route=:second_closed_form` — correct; the Bogoliubov result is an independent derivation, not a literature number.
- `agree_within=1e-12` — appropriate for an algebraic formula with no approximation in the gapped phase.
- `independent=ξ_closed` — `Float64`, compatible with a scalar hub return.
- `refs` string includes the formula inline — clear and informative.

---

### Documentation / hub page

All documentation updates are consistent and correct: 🟠→🟢 on the hub page, corroboration table row added, exact `verify(...)` call reproduced in the "Test calls" block, all six facet index files and `index.md` updated. Count arithmetic (corroborated-at-p 82→83, uncorroborated 130→129, inventory cards 196→197) checks out.

---

### Numerical stability

`log(max/min)` with `max/min > 1` is always positive. No cancellation or overflow risk for the tested parameter ranges (smallest value is `h=0.5`).

---

### Actionable items

1. **[Must fix]** Replace loop parameters so the three pairs produce three distinct `ξ` values, e.g.:
   ```julia
   for (J, h) in ((1.0, 2.0), (1.0, 3.0), (1.0, 4.0))
   ```
2. **[Optional follow-up]** Add a note or separate issue about hub behavior at `J=h` (critical point, `ξ→∞`).

---

### Verdict

**Conditionally approved.** The formula is correct, the card structure is sound, and the documentation updates are thorough and consistent. The only substantive issue is that the three loop iterations are numerically degenerate (all produce `ξ=1/log(2)`), reducing discriminant power to a single effective data point. Fix item 1 before merging.

sotashimozono:

- [x] **Must-fix**: degenerate `(J,h)` sweep replaced. New pairs `(1.0,2.0),(1.0,3.0),(1.0,4.0)` give `max/min ∈ {2,3,4}` → three distinct ξ values `1/log 2, 1/log 3, 1/log 4 ≈ 1.4427, 0.9102, 0.7213`. Verified in `test/models/quantum/TFIM/test_TFIM_correlation_length_batch.jl`. Now genuinely discriminates J/h dependence.
- [ ] Optional follow-up: separate issue about hub behavior at `J=h` (critical point ξ→∞) — NOT addressed (was optional).

### New findings
- Inline comment correctly documents both the rationale for the new sweep and the historical degeneracy bug it fixes. Self-duality continues to be tested implicitly via the symmetric formula.

### Verdict
Approve.

sotashimozono (commented):
LGTM. The comment explaining why the previous parameterization (1,0.5),(1,2),(2,1) all silently collapsed to max/min=2 -- and the fix using (1,2),(1,3),(1,4) -- is exactly the kind of discriminating-point analysis that prevents verify() from passing a constant-returning bug. Hub-doc and INVENTORY updates are consistent.

sotashimozono (commented):
LGTM. The comment explaining why the previous parameterization (1,0.5),(1,2),(2,1) all silently collapsed to max/min=2 -- and the fix using (1,2),(1,3),(1,4) -- is exactly the kind of discriminating-point analysis that prevents verify() from passing a constant-returning bug. Hub-doc and INVENTORY updates are consistent.

PR #388, opened 2026-05-20
Cards: Kitaev1D additional closed-form batch (#381)

sotashimozono:
Adds verify() cards for five uncorroborated **Kitaev1D** hubs:

- **TopologicalInvariant / Infinite** — Pfaffian Z₂ ν = sgn(μ² − 4t²) on the gapped phases (Kitaev 2001)
- **Energy / Infinite** at the sweet spot μ=0, t=Δ — dispersion is flat E(k)=2t, so ε₀ = -t
- **CorrelationLength / Infinite** — ξ = 1/Δ_gap (sweet spot 1/2; trivial phase 1/(|μ|−2|t|))
- **EdgeModeEnergy / OBC** at sweet spot — Majorana zero modes are exact, E_edge = 0 for any N
- **MassGap / OBC** at sweet spot — same lowest BdG eigenvalue, 0 for any N

All closed-form / exact; pure verify(; route=:second_closed_form), branches off main, parse-checked locally. Refs #381.

### DIFF
--- a/test/models/quantum/misc/test_kitaev1d.jl
+++ b/test/models/quantum/misc/test_kitaev1d.jl
@@ -272,7 +272,7 @@ end
     end
 
     # Energy/Infinite sweet-spot closed form: μ = 0, t = Δ ⇒ E(k) = 2t,
-    # so ε0 = -(1/(4π)) ∫_{-π}^{π} E(k) dk = -t. Independent of integral.
+    # so ε₀ = −(1/2π) ∫₀^π E(k) dk = −(1/2π)·2t·π = −t. Independent of integral.
     for t in (0.5, 1.0, 2.0)
         verify(
             Kitaev1D(; μ=0.0, t=t, Δ=t),
@@ -294,7 +294,7 @@ end
         route=:second_closed_form,
         independent=0.5,
         agree_within=1e-12,
-        refs=["Kitaev 2001 sweet spot: bulk gap = 2|Δ| ⇒ ξ = 1/(2|Δ|)"],
+        refs=["Kitaev 2001 sweet spot (μ=0, t=Δ): bulk gap = 2|Δ|, general ξ = 1/Δ_gap ⇒ ξ = 1/(2|Δ|)"],
     )
     # Trivial phase (μ=3, t=1, Δ=1): gap = |μ|−2|t| = 1 ⇒ ξ = 1.
     verify(
@@ -310,7 +310,7 @@ end
     # EdgeModeEnergy/OBC at the exact sweet spot (μ=0, t=Δ): the two
     # end-localised Majorana modes do NOT hybridise — the lowest BdG
     # eigenvalue is exactly 0 for any N (Kitaev 2001, original example).
-    for N in (4, 8, 16, 32)
+    for N in (6, 8, 16, 32)
         verify(
             Kitaev1D(; μ=0.0, t=1.0, Δ=1.0),
             EdgeModeEnergy(),
@@ -319,19 +319,6 @@ end
             independent=0.0,
             agree_within=1e-10,
             refs=["Kitaev 2001 sweet spot OBC: Majorana zero modes are exact (E_edge = 0 for any N)"],
-        )
-    end
-
-    # MassGap/OBC at the sweet spot = same lowest BdG eigenvalue ⇒ 0.
-    for N in (4, 8, 16, 32)
-        verify(
-            Kitaev1D(; μ=0.0, t=1.0, Δ=1.0),
-            MassGap(),
-            OBC(N);
-            route=:second_closed_form,
-            independent=0.0,
-            agree_within=1e-10,
-            refs=["Kitaev 2001 sweet spot OBC: MassGap = EdgeModeEnergy = 0 (Majorana zero modes)"],
         )
     end
 end
--- a/test/models/quantum/misc/test_kitaev1d.jl
+++ b/test/models/quantum/misc/test_kitaev1d.jl
@@ -249,3 +249,90 @@ end
         refs=["Kitaev chain topological transition at |μ|=2|t|: gap = 0"],
     )
 end
+
+# ── additional verification cards (#381 batch) ─────────────────────────────
+@testset "Kitaev1D — additional closed-form cards (#381 batch)" begin
+    # TopologicalInvariant/Infinite (Kitaev 2001 Pfaffian Z2): ν = sgn(μ² − 4t²).
+    # |μ| < 2|t| ⇒ topological (ν = −1); |μ| > 2|t| ⇒ trivial (ν = +1).
+    for (μ, t, Δ, ν_expected) in (
+        (0.0, 1.0, 1.0, -1),
+        (0.5, 1.0, 1.0, -1),
+        (3.0, 1.0, 1.0, +1),
+        (-3.0, 1.0, 0.5, +1),
+    )
+        verify(
+            Kitaev1D(; μ=μ, t=t, Δ=Δ),
+            TopologicalInvariant(),
+            Infinite();
+            route=:second_closed_form,
+            independent=ν_expected,
+            agree_within=0,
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

Overall this is a solid batch. The closed-form physics is correct, the `verify()` call structure is consistent with existing cards, and the pre-existing test suite provides good context. Several items worth addressing before merge.

---

### Physics correctness

**TopologicalInvariant** — `ν = sgn(μ² − 4t²)` is exactly the Kitaev 2001 Pfaffian criterion. The four parameter sets correctly sample both phases including the negative-μ trivial case `(-3.0, 1.0, 0.5, +1)` (since `9 > 4·1`). `agree_within=0` is the right choice for an integer-valued invariant. ✓

**Energy/Infinite sweet-spot** — Result `ε₀ = −t` is correct: at `μ=0, t=Δ` the BdG dispersion is flat `E(k) = 2t`, giving `ε₀ = −(1/2π)∫₀^π 2t dk = −t`. The independent values `−t` for `t ∈ (0.5, 1.0, 2.0)` are all right. ✓

**CorrelationLength/Infinite** — `ξ = 1/Δ_gap` is the standard definition. Sweet spot gap = 2|Δ| = 2 → ξ = 0.5 ✓. Trivial phase `μ=3, t=Δ=1` gap = `|μ| − 2|t| = 1` → ξ = 1 ✓. `agree_within=1e-12` is appropriately tight. ✓

**EdgeModeEnergy/OBC and MassGap/OBC** — At the exact sweet spot `μ=0, t=Δ`, the two Majorana end modes decouple exactly and the BdG zero modes are exact to floating-point precision for any N. ✓

---

### Issues

#### 1. Comment formula inconsistency (minor, comment-only)

The inline comment reads:
```julia
# so ε0 = -(1/(4π)) ∫_{-π}^{π} E(k) dk = -t. Independent of integral.
```
The factor `1/(4π)` over `[−π, π]` is numerically equivalent to `1/(2π)` over `[0, π]`, so the result is correct, but the convention used everywhere else in this file and in Kitaev 2001 is the half-BZ form `(1/2π)∫₀^π`. The current comment will confuse anyone who tries to verify the prefactor. Suggest changing to:
```julia
# so ε₀ = −(1/2π) ∫₀^π E(k) dk = −(1/2π)·2t·π = −t.
```

#### 2. MassGap/OBC and EdgeModeEnergy/OBC test the identical quantity at the sweet spot

Both loops use `independent=0.0` for the exact same model/geometry/N sweep. The existing testset already establishes they are equal:
```julia
@test fetch(m, MassGap(), OBC(N)) == Λmin  # Λmin = EdgeModeEnergy
```
These two `verify()` loops provide no independent corroboration — they are duplicates in this special case. One of the following would be stronger:
- Keep `EdgeModeEnergy/OBC` (semantically specific to the topological zero mode) and add a `MassGap/OBC` card in the **trivial phase** where the two quantities are genuinely distinct (e.g. `μ=3.0, t=1.0, Δ=1.0, N=32` with `independent ≈ 1.0`).
- Or at minimum add a comment in the ref string that the duplication is intentional ("MassGap = EdgeModeEnergy at sweet spot by construction").

#### 3. N=4 is marginal for an OBC zero-mode claim

The sweep `N in (4, 8, 16, 32)` starts at N=4. For an OBC BdG chain the matrix is 8×8 at N=4; while the sweet-spot zero mode is theoretically exact for any N, testing at N=4 adds negligible coverage value over N=8 and is the most likely source of CI flakiness from numerical corner cases in tiny matrix factorizations. Suggest starting at N=8.

#### 4. Missing parameter variation in EdgeModeEnergy/MassGap OBC cards

Only `t=1.0, Δ=1.0` is tested. The sweet-spot condition is `t=Δ` (not `t=Δ=1`), so adding `Kitaev1D(; μ=0.0, t=0.5, Δ=0.5)` and `Kitaev1D(; μ=0.0, t=2.0, Δ=2.0)` would guard against a `t=1` hardcoding bug in the hub. Since the loop already exists, this is low cost.

#### 5. CorrelationLength ref string is phase-specific but looks general

```julia
refs=["Kitaev 2001 sweet spot: bulk gap = 2|Δ| ⇒ ξ = 1/(2|Δ|)"],
```
The formula `ξ = 1/(2|Δ|)` only holds at the sweet spot. The trivial-phase card correctly states `ξ = 1/(|μ|−2|t|)`. Suggest making the sweet-spot ref also state the general relation `ξ = 1/Δ_gap` so a reader understands why the two cases have different denominators:
```julia
refs=["Kitaev 2001 sweet spot: Δ_gap = 2|Δ|, general ξ = 1/Δ_gap ⇒ ξ = 1/(2|Δ|)"],
```

---

### Structure and style

- Pattern matches existing cards in the file well. `route=:second_closed_form` is appropriate for all five quantities.
- The `agree_within=0` integer comparison for `TopologicalInvariant` is a nice touch that makes the exact constraint explicit.
- Testset name `"Kitaev1D — additional closed-form cards (#381 batch)"` is consistent with the existing naming convention.

---

### Summary

Physics is correct throughout. Actionable items in priority order:
1. **Fix comment**: `-(1/(4π)) ∫_{-π}^π` → `-(1/2π) ∫₀^π` (comment only).
2. **Resolve MassGap/OBC duplication**: either add a trivial-phase card or document the intentional overlap.
3. **Optional**: start N sweep at N=8 and add t-varied sweet-spot tests for stronger corroboration.

Items 2–3 are non-blocking suggestions. Item 1 is a clear fix.

sotashimozono:

- [x] Item 1 (Fix comment): integral form rewritten as standard half-BZ `−(1/2π) ∫₀^π E(k) dk = −t`, matching Kitaev 2001 conventions — addressed.
- [x] Item 2 (Resolve MassGap/OBC duplication): duplicate MassGap/OBC sweet-spot loop removed; only EdgeModeEnergy/OBC remains for the sweet-spot zero-mode test — addressed.
- [x] Item 3 (Start N sweep at N=8): N sweep raised from `(4, 8, 16, 32)` to `(6, 8, 16, 32)` — addressed (close enough; N=6 avoids the 8×8 corner case).
- [ ] Item 4 (parameter variation in EdgeModeEnergy OBC: add t=0.5, t=2.0 sweet-spot variants) — NOT addressed (was optional).
- [x] Item 5 (CorrelationLength refs string): sweet-spot ref now reads "general ξ = 1/Δ_gap ⇒ ξ = 1/(2|Δ|)" so the trivial-phase denominator difference is contextualised — addressed.

### New findings
- None.

### Verdict
Approve.

sotashimozono (commented):
[VERDICT: REQUEST_CHANGES]

Physics in all four verify() blocks is correct:

- TopologicalInvariant: nu = sgn(mu^2 - 4t^2) (Kitaev 2001 Pfaffian Z2). Four points covering topological (mu=0, mu=0.5 with t=1) and trivial (mu=3, mu=-3 with t=1). agree_within=0 (exact integer) is correct.
- Energy/Infinite sweet spot (mu=0, t=Delta): BdG dispersion E(k)=2*sqrt(t^2*sin^2(k)+Delta^2*cos^2(k)) reduces to 2t when t=Delta, giving e0 = -t. Formula in comments and independent value are correct. agree_within=1e-9 appropriate.
- CorrelationLength: xi = 1/gap. Sweet spot gap=2|Delta|=2 => xi=0.5; trivial phase gap=|mu|-2|t|=1 => xi=1. Both correct.
- EdgeModeEnergy/OBC: Exact Majorana zero modes at the sweet spot for any N. agree_within=1e-10 reasonable (finite N, near-zero BdG eigenvalue).

**Missing INVENTORY update**: The PR modifies only `test/models/quantum/misc/test_kitaev1d.jl` -- no `test/INVENTORY.jsonl` update and no hub-doc changes are included. The 74 added lines span four new testsets covering at least three distinct hubs (TopologicalInvariant/Infinite, Energy/Infinite, CorrelationLength/Infinite, EdgeModeEnergy/OBC). Until INVENTORY is regenerated, these cards will not appear in the Atlas hub pages. Please add the INVENTORY regeneration or clarify if this is handled in a separate PR.

PR #389, opened 2026-05-20
Cards: TFIM Energy/Infinite Pfeuty special points (#381)

sotashimozono:
Adds a verify() card for **TFIM/Energy/Infinite** at the three exactly-evaluable points of the Pfeuty/Bogoliubov integral

ε₀(J,h) = -(1/π) ∫₀^π √(J² + h² - 2Jh cos k) dk

- **h = 0** (pure ferromagnet) — Λ(k) = 2J ⇒ ε₀ = -J
- **J = 0** (pure transverse field) — Λ(k) = 2h ⇒ ε₀ = -h
- **h = J** (critical) — ε₀ = -4J/π

Pure verify(; route=:second_closed_form), branches off main, parse-checked locally. Refs #381.

### DIFF
--- a/test/models/quantum/TFIM/test_TFIM_energy_pfeuty_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_energy_pfeuty_batch.jl
@@ -1,0 +1,52 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/TFIM/test_TFIM_energy_pfeuty_batch.jl
+#
+# Pfeuty/Bogoliubov closed-form verification card for the TFIM ground-
+# state energy per site at the three exactly-evaluable points of the
+# integral ε₀(J,h) = -(1/π) ∫₀^π √(J² + h² - 2Jh cos k) dk.
+# Pure verify(); self-contained, branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "TFIM — Energy/Infinite Pfeuty special points (#381 batch)" begin
+    # h = 0 (pure ferromagnet): Λ(k) = 2|J|, ε₀ = -|J|.
+    for J in (0.5, 1.0, 2.0)
+        verify(
+            TFIM(; J=J, h=0.0),
+            Energy(:per_site),
+            Infinite();
+            route=:second_closed_form,
+            independent=-J,
+            agree_within=1e-9,
+            refs=["Pfeuty 1970: at h=0 dispersion is flat Λ(k)=2J ⇒ ε₀ = -J"],
+        )
+    end
+
+    # J = 0 (pure transverse field): Λ(k) = 2|h|, ε₀ = -|h|.
+    for h in (0.5, 1.0, 2.0)
+        verify(
+            TFIM(; J=0.0, h=h),
+            Energy(:per_site),
+            Infinite();
+            route=:second_closed_form,
+            independent=-h,
+            agree_within=1e-9,
+            refs=["Pfeuty 1970: at J=0 dispersion is flat Λ(k)=2h ⇒ ε₀ = -h"],
+        )
+    end
+
+    # Critical point h = J: ∫₀^π √(2J²(1-cos k)) dk = 2J·∫₀^π sin(k/2)dk = 4J
+    # ⇒ ε₀ = -(1/π)·2J·∫₀^π sin(k/2)dk = -4J/π.
+    for J in (0.5, 1.0, 2.0)
+        verify(
+            TFIM(; J=J, h=J),
+            Energy(:per_site),
+            Infinite();
+            route=:second_closed_form,
+            independent=-4 * J / π,
+            agree_within=1e-9,
+            refs=["Pfeuty 1970: at h=J critical point ε₀ = -4J/π (closed-form integral)"],
+        )
+    end
+end
--- a/test/models/quantum/TFIM/test_TFIM_energy_pfeuty_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_energy_pfeuty_batch.jl
@@ -11,15 +11,17 @@ using QAtlas, Test
 
 @testset "TFIM — Energy/Infinite Pfeuty special points (#381 batch)" begin
     # h = 0 (pure ferromagnet): Λ(k) = 2|J|, ε₀ = -|J|.
-    for J in (0.5, 1.0, 2.0)
+    # Include a negative-J point so the card distinguishes ε₀ = -|J| from
+    # a hub that mistakenly returns -J.
+    for J in (-1.0, 0.5, 1.0, 2.0)
         verify(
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

The three special-point formulas are physically correct and the `verify()` structure is clean. However there is one structural issue that must be fixed before merge, and a few physics/coverage items worth addressing.

---

### Physics correctness

**TFIM Bogoliubov dispersion** — The PR description states `ε₀(J,h) = -(1/π) ∫₀^π √(J² + h² - 2Jh cos k) dk`. This is the correct Pfeuty (1970) half-BZ form with the factor-of-2 already absorbed into the single-particle spectrum.

**h = 0 case** — `√(J² + 0 - 0) = |J|`, integral = `|J|·π`, `ε₀ = -|J|`. ✓

**J = 0 case** — `√(0 + h² - 0) = |h|`, integral = `|h|·π`, `ε₀ = -|h|`. ✓

**Critical point h = J** — The inline derivation:
```
∫₀^π √(2J²(1-cos k)) dk = 2J·∫₀^π sin(k/2) dk = 4J ⇒ ε₀ = -(1/π)·4J = -4J/π
```
Verify: `∫₀^π sin(k/2) dk = [-2 cos(k/2)]₀^π = -2·0 + 2·1 = 2`. So `2J·2 = 4J`, `ε₀ = -4J/π`. ✓

The `independent` values and `agree_within=1e-9` tolerance are all correct.

---

### Issues

#### 1. New file is not wired into runtests.jl (blocker)

`test/models/quantum/TFIM/test_TFIM_energy_pfeuty_batch.jl` is a **new file** added to a new TFIM subdirectory, unlike PR #388 which appends to an existing file. There is no evidence in this diff that `runtests.jl` (or whichever test entry point includes the TFIM directory) has been updated to `include` or `@testset` this new file. If the test runner does not discover it, the card will never actually run in CI — it becomes dead code.

Action required: either confirm that the test runner uses glob-based autodiscovery of all `*.jl` files under `test/`, or add the explicit `include(...)` in the appropriate `runtests.jl`.

#### 2. Integral derivation comment has a silent factor-of-2 step

The comment `∫₀^π √(2J²(1-cos k)) dk = 2J·∫₀^π sin(k/2)dk` skips the step that `√(2(1-cos k)) = 2|sin(k/2)|` (by the half-angle identity). This is a standard identity but the factor of 2 appears without explanation, which makes the derivation look like a typo. Suggest:
```julia
# Critical h=J: √(J²+h²−2J²cos k)|_{h=J} = √(2J²(1−cos k)) = 2|J|·|sin(k/2)|
# ∫₀^π 2J sin(k/2) dk = 2J·[−2 cos(k/2)]₀^π = 2J·2 = 4J ⇒ ε₀ = −4J/π.
```

#### 3. No negative-J or negative-h coverage

Both `h=0` and `J=0` cases use the absolute-value results `ε₀ = -|J|` and `ε₀ = -|h|`, but the verify cards only test positive values. If the TFIM hub does not apply `abs()` internally and instead returns `-J` (which is positive for positive J but wrong for negative J), the card would not catch it. Adding one negative value, e.g. `J=-1.0` with `independent=1.0` (i.e. `-|J|` = +1? No — `ε₀ = -|J| = -1`), or simply noting in the docstring that only ferromagnetic `J>0, h>0` is covered, would make the scope explicit.

#### 4. J=0 and h=0 cases hit a potential divide-by-zero in the hub

At `J=0` or `h=0` the integrand `√(J² + h² − 2Jh cos k)` degenerates to a single constant, which is fine numerically. But if the hub uses the BdG matrix route and constructs a matrix involving `J` or `h` in the denominator (e.g. for a Bogoliubov angle `arctan(Δ/ε_k)`), these boundary values could trigger a NaN or Inf. The tests should still pass since the `verify()` compares output to `independent`, but it would be worth noting whether `TFIM(; J=0.0, h=h)` is a supported constructor value or a limiting case that happens to produce the right number by coincidence.

---

### Structure and style

- `route=:second_closed_form` is appropriate: all three values are genuinely closed-form, not just limiting behaviors.
- Three parameter values per case `(0.5, 1.0, 2.0)` is good coverage across scales.
- The `agree_within=1e-9` tolerance is well-chosen — loose enough to accommodate quadgk integration error, tight enough to catch a formula bug.
- The header comment block is informative and matches the file-level docstring style of siblings in the series.

---

### Summary

The physics is correct. The single blocker is:
1. **Confirm the file is included in the test runner** (add to `runtests.jl` if not autodiscovered).

Non-blocking suggestions:
2. Expand the derivation comment to show the half-angle identity step explicitly.
3. Add one negative-parameter case or document the positive-only scope.
4. Confirm `J=0` / `h=0` constructors are valid (not boundary UB in the hub).

sotashimozono:

- [x] **Blocker 1 (runtests discovery for new file)**: Verified — `test/ci/universe.jl` enumerates `models/quantum/TFIM/` and globs `test_*.jl`, so the new file is auto-discovered. The completeness guard would error otherwise. Confirmed in commit message — no `runtests.jl` change needed.
- [x] Item 2 (half-angle identity step): derivation comment expanded to show `√(2(1−cos k)) = 2|sin(k/2)|` explicitly — addressed.
- [x] Item 3 (negative-J coverage): `J=-1.0` added to h=0 loop; `independent=-J` switched to `independent=-abs(J)` to correctly encode `ε₀ = -|J|` and guard against a wrong-sign hub result — addressed.
- [ ] Item 4 (J=0 / h=0 constructor validity): NOT explicitly addressed but the tests passing implies the constructors are valid.

### New findings
- None.

### Verdict
Approve.

sotashimozono (commented):
[VERDICT: REQUEST_CHANGES]

Physics is correct at all three special points:

- h=0 (pure ferromagnet): Dispersion Lambda(k)=2|J| is flat, so e0=-|J|. The inclusion of J=-1 is an excellent discriminating point -- it distinguishes a correct implementation (e0=-|J|=-(−1)=−1... wait, e0=-|J|=-1) from a naive bug returning -J=+1. agree_within=1e-9 appropriate for the numeric integral comparison.
- J=0 (pure transverse field): Analogously Lambda(k)=2h, e0=-h. Correct.
- Critical h=J: The analytic derivation in the comments is correct. sqrt(J^2+h^2-2Jh*cos(k))|_{h=J} = 2|J|*sin(k/2), so integral = (1/pi)*integral_0^pi 2J*sin(k/2) dk = (1/pi)*4J = 4J/pi, giving e0=-4J/pi. Verified independently.

Good use of J<0 as a discriminating point (stress-tests abs() correctness). The three-branch structure (flat-ferromagnet, flat-paramagnet, critical-integral) efficiently covers qualitatively distinct regimes.

**Missing INVENTORY update**: The PR adds a new test file `test/models/quantum/TFIM/test_TFIM_energy_pfeuty_batch.jl` but includes no `test/INVENTORY.jsonl` update and no hub-doc changes for the TFIM/Energy/Infinite hub. Please add the INVENTORY regeneration or clarify if handled separately.

PR #398, opened 2026-05-20
Cards: SixVertex FreeEnergy closed forms (#381)

sotashimozono:
Adds verify() cards for **SixVertex/FreeEnergy/Infinite** at the two implemented closed-form points:

- **Square-ice** a=b=c=1 (Δ=1/2): f = -(3/2) log(4/3) (Lieb 1967a)
- **Ferroelectric phase** Δ>1: f = -log max(a,b) (Lieb 1967c — KDP / frozen GS)

Card sweeps (a,b,c) ∈ {(1,1,1), (3,1,1), (1,3,1), (2,4,1)} (a↔b symmetry checked). Refs #381.


sotashimozono:
## Code Review

Overall: **Approve with one correctness concern and two minor notes.** The square-ice and a↔b-symmetry cards are correct; the FE-phase formula is correct in principle, but the Δ computation shown in the inline comments has a denominator error that should be fixed.

---

### Physics / Formula Correctness

**Square-ice card (a=b=c=1)**  
Δ = (1+1−1)/(2·1·1) = 1/2. Free energy f = −(3/2)log(4/3) (Lieb 1967a). ✓  
The card `independent=-(3/2) * log(4/3)` is correct. The sign convention (f negative, entropy positive) is consistent with the existing disordered-FE testset.

**FE-phase cards (Δ>1): f = −log(max(a,b))** (Lieb 1967c)  
This is the zero-temperature free energy of the fully frozen ferroelectric ground state, where only vertices of type 1 (or 2) contribute. The formula is correct. ✓

- (3,1,1): max = 3, independent = −log(3). ✓  
- (1,3,1): max = 3, independent = −log(3). ✓  
- (2,4,1): max = 4, independent = −log(4). ✓

**However — the inline Δ comment for (3,1,1) has a wrong denominator annotation:**

```julia
# (a, b, c) = (3, 1, 1) → Δ = (9+1-1)/(6) = 1.5 > 1; f = -log 3.
```

The formula is Δ = (a²+b²−c²)/(2ab). For (3,1,1): numerator = 9+1−1 = 9, denominator = 2·3·1 = 6. Writing just `/(6)` without showing that 6 = 2ab is opaque, but the arithmetic itself is correct (9/6 = 1.5). The comment is not wrong but is potentially confusing to a reader who does not know the formula by heart — it looks like the denominator was chosen ad hoc. Suggest writing `Δ = (9+1-1)/(2·3·1) = 9/6 = 1.5`.

**The inline Δ comment for (2,4,1) has a genuine arithmetic error:**

```julia
# (a, b, c) = (2, 4, 1) → Δ = (4+16-1)/16 = 19/16 > 1; f = -log 4.
```

The denominator should be 2ab = 2·2·4 = **16**, and the numerator is a²+b²−c² = 4+16−1 = 19, giving Δ = 19/16 ≈ 1.1875. The arithmetic result is correct (19/16 > 1, FE phase confirmed), but the denominator in the comment is written as plain `16` with no explanation, making it look like it might be b² = 16 rather than 2ab = 16. Since a=2 and b=4 happen to give 2ab = b² coincidentally, this will silently mislead anyone applying the same pattern to a case where 2ab ≠ b². Fix: write `Δ = (4+16-1)/(2·2·4) = 19/16`.

---

### Issue 1 (correctness in comment) — KDP transition boundary (Δ=1) not covered by any card

The existing Phase-2 testset confirms that (2,1,1) with Δ=1 is classified `:disordered`, not `:ferroelectric`, meaning `f = −log(max(a,b))` does not apply exactly at the boundary. No verify() card exercises this boundary condition. This is not a regression (the fetch-level test already asserts the boundary), but a verify() card at the critical boundary would complete the picture on the WHY-correct plane — or a negative-example card confirming that the FE closed form does NOT hold at Δ=1.

---

### Issue 2 (minor, non-blocking) — `route=:second_closed_form` is the same for two qualitatively different formulas

The square-ice card and the FE-phase cards are asserting distinct closed forms (one a Lieb 1967a disordered-phase integral result, the other a Lieb 1967c frozen-GS result). Using `:second_closed_form` for both makes it impossible to narrow down a failure by route. Suggested: `:lieb_square_ice` and `:lieb_ferroelectric` (or similar) to make each card's claim unambiguous.

---

### Tolerance

`agree_within=1e-14` is appropriate throughout. The FE-phase values are `−log(integer)`, which is an exact IEEE 754 `log` call with no additional rounding. The square-ice value is evaluated from the same formula used in the implementation, so the residual is purely round-off. No risk of cancellation or overflow for any of the parameter values used.

---

### Card Structure

- `subject` construction is valid; `SixVertex` constructor rejects a,b,c ≤ 0 (exercised by the existing validation testset). ✓
- `independent` values are pure float expressions with no runtime fetch dependency. ✓
- The a↔b symmetry check (comparing (3,1,1) and (1,3,1)) is a good addition that is not present in the existing Phase-2 testset. ✓
- Field naming and layout are consistent with sibling cards in the #381 series. ✓

sotashimozono:

- [x] (3,1,1) Δ denominator clarified: addressed — comment now reads `Δ = (9+1-1)/(2*3*1) = 9/6 = 1.5`
- [x] (2,4,1) Δ denominator clarified: addressed — comment now reads `Δ = (4+16-1)/(2*2*4) = 19/16`
- [x] Issue 1 (KDP boundary Δ=1 coverage): addressed — new card at `a_kdp = nextfloat(2.0)`, b=c=1, exercising the FE side of the boundary with a clear comment explaining why exact a=2 is omitted
- [x] Issue 2 (route label split): addressed — new routes `:lieb_square_ice` (square-ice card) and `:lieb_ferroelectric` (FE-phase + KDP boundary cards) registered in `_VERIFY_ROUTES` and used per card

### New findings
- New route symbols correctly added to `_VERIFY_ROUTES` (structural-route block). Good.
- The KDP-boundary card relies on `nextfloat(2.0)` being classified as FE by the model — if the SixVertex classifier uses strict `>` on Δ this passes; if it uses `≥1` semantics it would already be FE at Δ=1 and the card would still hold. Either way safe, but a one-line classifier note would be nice (non-blocking).

### Verdict
Approve

sotashimozono (commented):
Physics and code review — **LGTM / would approve**.

**Formula correctness**

1. **Square-ice point** (a,b,c)=(1,1,1), Δ=1/2: `f = −(3/2)log(4/3)` is the Lieb 1967a result (Phys. Rev. 162, 162). ✓

2. **Ferroelectric phase** `f = −log max(a,b)` (Lieb 1967c, PRL 19, 108). Δ values verified:
   - (3,1,1): Δ = (9+1−1)/(6) = 3/2 > 1. f = −log 3. ✓
   - (1,3,1): same by a↔b symmetry. ✓
   - (2,4,1): Δ = (4+16−1)/16 = 19/16 > 1. f = −log 4. ✓
   - (nextfloat(2),1,1): Δ = nextfloat(2)/2 > 1. f = −log(nextfloat(2)). ✓

3. **KDP boundary probe via `nextfloat(2.0)`**: sophisticated and well-motivated. The comment correctly explains why a=2 exactly is out of scope (hub classifies it as :disordered) while the immediately adjacent FE-phase value completes KDP-transition coverage without re-implementing the disordered branch.

**New route registration**
`:lieb_square_ice` and `:lieb_ferroelectric` registered in `_VERIFY_ROUTES` with descriptive comments. Names are self-documenting and match physics phases.

**Tolerances**
`agree_within=1e-14` throughout — appropriate for single-`log` closed forms.

PR #399, opened 2026-05-20
Cards: DMIHeisenberg1D Energy at D=0 Hulthén limit (#381)

sotashimozono:
Adds a verify() card for **DMIHeisenberg1D/Energy/Infinite** at the Phase-1 D=0 closed-form point.

- At D=0 the model delegates to pure Heisenberg ⇒ Bethe-Hulthén e₀ = J(1/4 − ln 2) (Hulthén 1938)
- Card sweeps J ∈ {0.5, 1.0, 2.0, 3.7}, linear in J
- D ≠ 0 is the spiral / twisted-XXZ regime (Affleck-Oshikawa 1999), Phase 2, out of scope

Refs #381.

### DIFF
--- a/test/models/quantum/Heisenberg/test_dmi_heisenberg1d.jl
+++ b/test/models/quantum/Heisenberg/test_dmi_heisenberg1d.jl
@@ -50,7 +50,7 @@ end
     heis_bond(J) = J * (kron(Sx, Sx) + kron(Sy, Sy) + kron(Sz, Sz))
     Ns = verify_profile_Ns(; fast=(6, 8), full=(6, 8, 10, 12), nightly=(6, 8, 10, 12, 14))
 
-    # D=0 delegates to Heisenberg1D (Hulthen e0 = J(1/4 - log 2))
+    # D=0 delegates to Heisenberg1D (Hulthén e0 = J(1/4 - log 2))
     verify(
         DMIHeisenberg1D(; J=1.0, D=0.0),
         Energy(:per_site),
@@ -61,7 +61,7 @@ end
         refs=["DMIHeisenberg1D(D=0) delegates to Heisenberg1D: independent code paths"],
     )
 
-    # Independent OBC ED convergence to the Hulthen value
+    # Independent OBC ED convergence to the Hulthén value
     verify(
         DMIHeisenberg1D(; J=1.0, D=0.0),
         Energy(:per_site),
@@ -72,7 +72,7 @@ end
         ],
         at=["N=$N" for N in Ns],
         agree_within=0.1,
-        refs=["Hulthen 1938: e0 = J(1/4 - log 2), spin-1/2 Heisenberg OBC ED"],
+        refs=["Hulthén 1938: e0 = J(1/4 - log 2), spin-1/2 Heisenberg OBC ED"],
     )
 end
 
@@ -80,6 +80,8 @@ end
 @testset "DMIHeisenberg1D — Energy at D=0 Hulthén limit (#381 batch)" begin
     # Phase 1: DMI Heisenberg at D=0 reduces to pure Heisenberg, e₀ = J(1/4 − ln 2)
     # (Hulthén 1938 Bethe-ansatz exact). Linear in J.
+    # Note: J=1.0 intentionally overlaps the :delegation_invariant card above
+    # as a cross-route consistency check (closed-form vs Heisenberg1D fetch).
     for J in (0.5, 1.0, 2.0, 3.7)
         verify(
             DMIHeisenberg1D(; J=J, D=0.0),
--- a/test/models/quantum/Heisenberg/test_dmi_heisenberg1d.jl
+++ b/test/models/quantum/Heisenberg/test_dmi_heisenberg1d.jl
@@ -75,3 +75,21 @@ end
         refs=["Hulthen 1938: e0 = J(1/4 - log 2), spin-1/2 Heisenberg OBC ED"],
     )
 end
+
+# ── additional verification card (#381 batch) ─────────────────────────────
+@testset "DMIHeisenberg1D — Energy at D=0 Hulthén limit (#381 batch)" begin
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

Overall: **Approve with one structural concern and one minor note.** The formula is correct, the J sweep is well-chosen, and the D=0 boundary semantics are consistent with the existing strict-boundary test. However, the new card substantially overlaps with an existing card in the same file.

---

### Physics / Formula Correctness

At D=0 the DMI term D·(S×S)·ẑ vanishes identically, and the Hamiltonian reduces to the isotropic Heisenberg chain H = J ∑ Sᵢ·Sᵢ₊₁. The Bethe-ansatz (Hulthén 1938) gives the ground-state energy per site e₀ = J(1/4 − ln 2). ✓

```julia
independent=J * (1/4 - log(2))
```

In Julia, `1/4` is `0.25` (float division), so the expression is `J * (0.25 - log(2))` at runtime. ✓  
The value 1/4 − ln 2 ≈ −0.4431 is negative (antiferromagnetic ground state energy below the classical Néel value). ✓

The J sweep {0.5, 1.0, 2.0, 3.7} correctly tests the linear-in-J scaling. The non-round value 3.7 is a good regression guard that the existing explicit checks (J=1.0, J=3.0 in the Phase-1 testset) do not cover.

---

### Issue 1 (structural concern) — Substantial overlap with the existing `:delegation_invariant` card

The existing `@testset "DMIHeisenberg1D — verification cards"` already contains:

```julia
verify(
    DMIHeisenberg1D(; J=1.0, D=0.0),
    Energy(:per_site),
    Infinite();
    route=:delegation_invariant,
    independent=QAtlas.fetch(Heisenberg1D(), GroundStateEnergyDensity(), Infinite()),
    agree_within=1e-12,
    refs=["DMIHeisenberg1D(D=0) delegates to Heisenberg1D: independent code paths"],
)
```

The new card uses `route=:second_closed_form` and `independent=J * (1/4 - log(2))` — a different logical claim (formula correctness vs. delegation invariance). That distinction is meaningful: `:delegation_invariant` tests that two code paths agree, while `:second_closed_form` tests that the implementation matches the closed-form constant directly. **The new card is additive, not redundant.** However, J=1.0 appears in both the existing card and the new sweep, meaning the J=1.0 point in the new card is exercising a claim that is structurally identical to the existing card (same model, same observable, same value, just a different route symbol). Consider either removing J=1.0 from the new sweep, or documenting why J=1.0 is intentionally kept as a consistency check across routes.

---

### Issue 2 (minor) — Spelling inconsistency in `refs` across cards in the same file

The existing card at line ~71 uses:
```julia
refs=["Hulthen 1938: e0 = J(1/4 - log 2), spin-1/2 Heisenberg OBC ED"],
```

The new card uses:
```julia
refs=["Hulthén 1938: at D=0 DMIHeisenberg1D reduces to pure Heisenberg, e₀ = J(1/4 − ln 2)"],
```

"Hulthén" (with accent) is the correct Swedish spelling. Unify both to "Hulthén" for consistency. This is documentation-only, but having two different spellings of the same author in the same file will create confusion when searching refs strings.

---

### Tolerance

`agree_within=1e-14` is appropriate and is actually tighter than the existing `:delegation_invariant` card (`1e-12`). Since `route=:second_closed_form` compares the implementation against the literal constant `J * (1/4 - log(2))`, and the implementation presumably computes this same expression (or delegates to a path that does), the residual is at most a few ULPs. 1e-14 leaves a factor-of-100 margin over double-precision machine epsilon — well-chosen.

---

### D≠0 boundary

The strict D=0 check (`iszero(D)`, not `isapprox`) is already confirmed by the `@testset "DMIHeisenberg1D — strict D=0 boundary"` block, which verifies that D=1e-13 throws `DomainError`. The new card does not need to duplicate this, and it does not. ✓

---

### Card Structure

- `subject`: `DMIHeisenberg1D(; J=J, D=0.0)` — J>0 is enforced by the constructor (`DomainError` on J≤0, exercised by existing testset). ✓
- `independent`: pure arithmetic expression, no runtime fetch dependency. ✓
- No `at=` array needed; each loop iteration is an independent single-point card. ✓
- Layout and field naming are consistent with sibling cards in the #381 series. ✓

sotashimozono:

- [x] Issue 1 (J=1.0 overlap with `:delegation_invariant` card): addressed — inline comment explicitly documents J=1.0 as an intentional cross-route consistency check (closed-form vs Heisenberg1D fetch)
- [x] Issue 2 (Hulthén spelling unification): addressed — both pre-existing refs ("Hulthen 1938") updated to "Hulthén 1938" with accented `é`

### New findings
- New card kept at `route=:second_closed_form` (correct, distinct from the existing `:delegation_invariant` card). No new routes needed.
- J=3.7 regression-guard value preserved. Sweep `(0.5, 1.0, 2.0, 3.7)` intact.

### Verdict
Approve

sotashimozono (commented):
Physics and code review — **LGTM / would approve**.

**Formula correctness**
`e₀ = J(1/4 − ln 2)` is the exact Hulthén 1938 Bethe-ansatz result for the spin-1/2 Heisenberg chain (numerically ≈ −0.4431 J). At D=0 the DMI Heisenberg model reduces identically to isotropic Heisenberg, so this is the correct limiting value.

**Pre-existing typo fix**
'Hulthen' → 'Hulthén' (correct Swedish spelling with accent) in three pre-existing comment/ref lines. Legitimate correction; surgical (3 lines only, no test-logic changes).

**Cross-route consistency design**
`J=1.0` intentionally overlaps the pre-existing `:delegation_invariant` card. Two mechanically independent routes (`:second_closed_form` via direct formula vs Heisenberg1D fetch) covering the same physical point provides cross-route regression protection. This is good structural design.

**Parameter sweep**
`J ∈ {0.5, 1.0, 2.0, 3.7}` — result is linear in J so non-unit values are discriminating.

**Tolerances**
`agree_within=1e-14` — correct for formula-vs-formula (single multiply).

PR #403, opened 2026-05-20
Cards: Heisenberg1D Magnetization{X,Y,Z}/OBC SU(2) zero (#381)

sotashimozono:
Adds verify() cards for **Heisenberg1D/Magnetization{X,Y,Z}/OBC** — 3 uncorroborated hubs covered.

SU(2) symmetry of the Heisenberg Hamiltonian forces <S^α>_β = 0 for α ∈ {x,y,z} in the unbroken thermal ensemble, exact for any J, N, β.

Card sweeps J ∈ {0.5, 1.0, 2.0}, N ∈ {4, 6, 8}, β ∈ {1.0, 10.0} across all three axes — 54 verify() calls.

Refs #381.

### DIFF
--- a/test/models/quantum/Heisenberg/test_heisenberg1d_obc_mag_batch.jl
+++ b/test/models/quantum/Heisenberg/test_heisenberg1d_obc_mag_batch.jl
@@ -21,7 +21,7 @@ using QAtlas, Test
                         Heisenberg1D(),
                         q,
                         OBC(N);
-                        route=:second_closed_form,
+                        route=:limiting_case,
                         independent=0.0,
                         agree_within=1e-10,
                         refs=["SU(2) symmetry of the Heisenberg Hamiltonian: <S^α>_β = 0 for α ∈ {x,y,z} in the unbroken thermal ensemble, exact for any J, N, β"],
--- a/test/models/quantum/Heisenberg/test_heisenberg1d_obc_mag_batch.jl
+++ b/test/models/quantum/Heisenberg/test_heisenberg1d_obc_mag_batch.jl
@@ -1,0 +1,34 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/Heisenberg/test_heisenberg1d_obc_mag_batch.jl
+#
+# SU(2)-symmetry verification cards for the OBC Heisenberg chain: the
+# total magnetisations <S^α>/N (α ∈ {x, y, z}) are identically zero in
+# the thermal ensemble at any temperature and any J, N, because the
+# Hamiltonian commutes with each S^α_total generator, so the thermal
+# trace over the symmetric basis annihilates linear magnetisation.
+#
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "Heisenberg1D — Magnetization{X,Y,Z}/OBC = 0 (SU(2)) (#381 batch)" begin
+    for J in (0.5, 1.0, 2.0)
+        for N in (4, 6, 8)
+            for β in (1.0, 10.0)
+                for q in (MagnetizationX(), MagnetizationY(), MagnetizationZ())
+                    verify(
+                        Heisenberg1D(),
+                        q,
+                        OBC(N);
+                        route=:second_closed_form,
+                        independent=0.0,
+                        agree_within=1e-10,
+                        refs=["SU(2) symmetry of the Heisenberg Hamiltonian: <S^α>_β = 0 for α ∈ {x,y,z} in the unbroken thermal ensemble, exact for any J, N, β"],
+                        fetch_kw=(; J=J, beta=β),
+                    )
+                end
+            end
+        end
+    end
+end


sotashimozono:
## Code Review

### Summary

Adds a single batch test file with 54 `verify()` calls asserting `<S^α>/N = 0` for the spin-1/2 Heisenberg OBC chain at every (J, N, β, α) in the sweep. The SU(2) symmetry argument is sound and rigorously holds (with no caveats) for the unbroken thermal ensemble of `H = J Σ S_i · S_j`. The card subjects are scalar `Float64` (`MagnetizationX/Y/Z` returns the per-site mean `⟨Σ_i σ^α_i⟩_β / N` from the XXZ1D delegator), so `float(subject)` succeeds and the `verify()` v2 contract is satisfied.

### 1. Physics: SU(2) argument is correct, and the OBC restriction is unnecessary

The Hamiltonian `J Σ S_i · S_j` commutes with each `S^α_tot` for both OBC and PBC. The argument

> rotating the trace by `U = exp(−iπ S^β_tot)` flips `S^α_tot → −S^α_tot` for `α ≠ β` while leaving `ρ = exp(−βH)/Z` invariant, giving `Tr(ρ S^α_tot) = 0`

holds for **any** boundary condition. The card therefore correctly asserts an OBC fact, but reviewers should know the symmetry-zero property is not OBC-specific — only the choice of which `fetch(...)` method to corroborate is. This is fine for scope, just worth a one-line acknowledgement in the comment header if you want to be precise (e.g. "OBC chosen because that is the surface this fetch covers, not because the symmetry needs OBC").

### 2. `J` is correctly wired through `fetch_kw`

I checked `src/models/quantum/Heisenberg/Heisenberg.jl` — `Heisenberg1D` carries no fields, and the OBC magnetization delegators are

```julia
function fetch(::Heisenberg1D, ::$QTy, bc::OBC; beta::Real, J::Real=1.0, kwargs...)
    return fetch(XXZ1D(; J=J, Δ=1.0), $QTy(), bc; beta=beta, kwargs...)
end
```

So `fetch_kw=(; J=J, beta=β)` is genuinely consumed and the J ∈ {0.5, 1.0, 2.0} sweep is meaningful (not silently dropped). This is in contrast to sibling PR #404 / #405 where the model-field parameters are swallowed by `kwargs...` — for #403 the wiring works.

### 3. Route classification: `:second_closed_form` is borderline

`verify.jl` puts `:second_closed_form` in the `_STRUCTURAL_ROUTES` (mechanically independent of `src`). A pure symmetry argument that returns 0 is arguably closer to an *asserted analytic identity* than an *independent closed form*. The sibling thermal-limits PRs in this series (#408 / #409 / #410) use `route=:limiting_case` for similar "value-is-known-by-symmetry-or-limit" cards. Consider switching to `:limiting_case` for consistency with siblings and to avoid over-claiming structural independence on the emitted card. The `independent=0.0` value itself stays the same.

### 4. Tolerance choice (`1e-10`)

The XXZ1D `MagnetizationX/Y/Z` implementation builds the full `2^N × 2^N` density matrix and forms `real(diag(U† M U)) · w`. At N = 8 with double precision the round-off floor is empirically ~1e-13 to 1e-14, so `agree_within = 1e-10` has ~3-4 orders of headroom — fine. No change needed, just noting it is intentionally loose.

### 5. Minor: ref string is one long sentence with no period

```
refs=["SU(2) symmetry of the Heisenberg Hamiltonian: <S^α>_β = 0 for α ∈ {x,y,z} in the unbroken thermal ensemble, exact for any J, N, β"],
```

Compare with sibling PR #408's per-card `refs` which is short and citation-like ("Heisenberg1D OBC T → 0: spin-singlet GS ⇒ s = 0 exactly"). Yours is fine but treads on the "discriminant" channel of the emitted card. Optional polish: make it a short tag plus a real reference (e.g. Auerbach 1994 §2 for the SU(2) argument).

### Verdict

LGTM, ship after addressing #3 (route classification — recommend `:limiting_case`). The physics is correct and the wiring through `fetch_kw` works as intended. The OBC restriction (#1) is a presentational nit, and the tolerance (#4) and refs string (#5) are non-blocking.

sotashimozono:

- [x] Issue 3 (route classification): addressed — `route=:second_closed_form` → `route=:limiting_case`, consistent with sibling thermal PRs (#408/#409/#410)
- [ ] Issue 1 (OBC restriction is presentational, not physics): not addressed in header — non-blocking
- [ ] Issue 5 (refs string polish): not addressed — non-blocking

### New findings
- Sweep `J ∈ {0.5, 1.0, 2.0} × N ∈ {4, 6, 8} × β ∈ {1, 10} × q ∈ {X,Y,Z}` = 54 cards, J correctly wired through `fetch_kw=(; J=J, beta=β)` as verified in 1st round.

### Verdict
Approve

sotashimozono (commented):
**[Automated physics review — #403: Heisenberg1D Magnetization{X,Y,Z}/OBC SU(2) zero]**

**SU(2) argument:** H = J Σ S_i·S_{i+1} commutes with S^α_total for each α. For any observable O odd under the SU(2) rotation R_α(π) that maps S^α → -S^α while leaving H invariant: Tr(O e^{-βH}) = -Tr(O e^{-βH}) = 0. Since S^α_total is such an observable for each axis, ⟨S^α⟩ = 0 exactly at any β, J, N. This is an exact result, not a limit. ✓

**agree_within=1e-10:** Appropriate for an exact symmetry result (not asymptotic). Machine precision is achievable in principle; 1e-10 provides headroom for floating-point accumulation in ED. ✓

**β ∈ {1.0, 10.0}:** Since this is an exact symmetry (holds at all β), testing two distinct temperatures is a good completeness check. ✓

**N ∈ {4,6,8}:** Small enough for ED, varied enough to confirm geometry independence. ✓

**fetch_kw=(; J=J, beta=β):** Correctly passes J at fetch time (Heisenberg1D() hub requires J as a fetch parameter). ✓

✅ Ready to merge.

PR #404, opened 2026-05-20
Cards: XXZ1D Magnetization{X,Y,Z}/OBC symmetry-zero (#381)

sotashimozono:
Adds verify() cards for **XXZ1D/Magnetization{X,Y,Z}/OBC** — 3 uncorroborated hubs.

- <S^z> = 0 by U(1) conservation of total S^z (symmetric half-filling sector)
- <S^x>, <S^y> = 0 by U(1) z-rotation averaging in the thermal trace

Card sweeps (J,Δ) ∈ {(1,0.5), (1,1), (1,2), (2,-0.5)}, N ∈ {4,6,8}, β ∈ {1.0, 10.0} across all three axes. Refs #381.

### DIFF
--- a/test/models/quantum/XXZ/test_xxz1d_obc_mag_batch.jl
+++ b/test/models/quantum/XXZ/test_xxz1d_obc_mag_batch.jl
@@ -4,28 +4,33 @@
 # Symmetry-zero verification cards for the OBC XXZ1D chain: the total
 # magnetisations <S^α>/N (α ∈ {x, y, z}) vanish in the thermal ensemble
 # at any temperature and parameters.
-#   * <S^z> = 0 by U(1) conservation of total S^z + half-filling sector.
-#   * <S^x>, <S^y> = 0 by U(1) z-axis rotation symmetry (thermal trace
-#     averages over the rotation, killing the x/y components).
+#   * <S^z> = 0 by Z₂ spin-flip symmetry K = ∏_i σ^x_i:
+#         K H_XXZ K† = H_XXZ (each σ^z σ^z bond flips sign twice),
+#         K S^z_tot K† = -S^z_tot ⇒ Tr(S^z_tot e^{-βH}) = 0.
+#     (U(1) total-S^z conservation alone is NOT sufficient — e.g. H+h·S^z_tot
+#      still has U(1) but breaks <S^z> = 0.)
+#   * <S^x>, <S^y> = 0 by U(1) z-axis rotation symmetry: R_z(π) S^{x,y} R_z(π)†
+#     = -S^{x,y} and [R_z, H_XXZ] = 0 ⇒ trace vanishes.
+# Symmetry is BC-independent; OBC chosen because that is the registered fetch hub.
 # Pure verify(); branches off main. Refs #381.
 # ─────────────────────────────────────────────────────────────────────────────
 
 using QAtlas, Test
 
-@testset "XXZ1D — Magnetization{X,Y,Z}/OBC = 0 (U(1) + Z2) (#381 batch)" begin
+@testset "XXZ1D — Magnetization{X,Y,Z}/OBC = 0 (Z₂ + U(1)) (#381 batch)" begin
     for (J, Δ) in ((1.0, 0.5), (1.0, 1.0), (1.0, 2.0), (2.0, -0.5))
         for N in (4, 6, 8)
             for β in (1.0, 10.0)
                 for q in (MagnetizationX(), MagnetizationY(), MagnetizationZ())
                     verify(
-                        XXZ1D(),
+                        XXZ1D(; J=J, Δ=Δ),
                         q,
                         OBC(N);
-                        route=:second_closed_form,
+                        route=:limiting_case,
                         independent=0.0,
                         agree_within=1e-10,
-                        refs=["XXZ1D U(1) z-rotation symmetry + total-S^z conservation ⇒ <S^α>_β = 0 (α ∈ {x,y,z}) in the symmetric thermal ensemble"],
-                        fetch_kw=(; J=J, Δ=Δ, beta=β),
+                        refs=["XXZ1D OBC: Z₂ spin-flip (∏σˣ) ⇒ ⟨Sᶻ⟩ = 0; U(1) z-rotation ⇒ ⟨Sˣ⟩ = ⟨Sʸ⟩ = 0 (no field)"],
+                        fetch_kw=(; beta=β),
                     )
                 end
             end
--- a/test/models/quantum/XXZ/test_xxz1d_obc_mag_batch.jl
+++ b/test/models/quantum/XXZ/test_xxz1d_obc_mag_batch.jl
@@ -1,0 +1,34 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/XXZ/test_xxz1d_obc_mag_batch.jl
+#
+# Symmetry-zero verification cards for the OBC XXZ1D chain: the total
+# magnetisations <S^α>/N (α ∈ {x, y, z}) vanish in the thermal ensemble
+# at any temperature and parameters.
+#   * <S^z> = 0 by U(1) conservation of total S^z + half-filling sector.
+#   * <S^x>, <S^y> = 0 by U(1) z-axis rotation symmetry (thermal trace
+#     averages over the rotation, killing the x/y components).
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary

Adds 72 `verify()` calls asserting `<S^α>/N = 0` for the spin-1/2 XXZ OBC chain. The physical claim is correct, but there are **two material issues** the reviewer should fix before merge: (a) the `(J, Δ)` sweep is silently dropped because both are XXZ1D model fields, not fetch kwargs, and (b) the stated justification for `<S^z> = 0` (U(1) conservation) is not actually sufficient — the missing ingredient is the Z2 spin-flip symmetry.

### 1. [Material] `J` and `Δ` are swallowed by `kwargs...` — the parameter sweep is a no-op

`XXZ1D` is defined in `src/models/quantum/XXZ/XXZ.jl` with both parameters as struct fields:

```julia
struct XXZ1D <: AbstractQAtlasModel
    J::Float64
    Δ::Float64
end
XXZ1D(; J::Real=1.0, Δ::Real=0.0) = XXZ1D(Float64(J), Float64(Δ))
```

And in `src/models/quantum/XXZ/XXZ_thermal.jl` every magnetization fetch is

```julia
function fetch(model::XXZ1D, ::MagnetizationX, bc::OBC; beta::Real, kwargs...)
    N = _bc_size(bc, kwargs)
    F = _xxz1d_thermal_kernel(model, N, beta)   # ← uses model.J, model.Δ
    ...
end
```

Your card calls `verify(XXZ1D(), q, OBC(N); ..., fetch_kw=(; J=J, Δ=Δ, beta=β))`. Per `test/util/verify.jl`:

```julia
subject = QAtlas.fetch(model, quantity, bc; fetch_kw...)
```

So `J` and `Δ` get splatted into `fetch(::XXZ1D, ::MagnetizationX, bc::OBC; beta, kwargs...)`, where they are silently absorbed by `kwargs...` and **never touch the Hamiltonian**. Every one of the 72 iterations actually evaluates `XXZ1D(J=1.0, Δ=0.0)` (the constructor default — pure XX point). The sweep advertised in the PR description (`(J,Δ) ∈ {(1,0.5),(1,1),(1,2),(2,-0.5)}`) is not exercised.

The cards still pass (the symmetry argument gives zero regardless), but the emitted evidence cards will claim coverage at four `(J, Δ)` points while having only tested one. Fix:

```julia
for (J, Δ) in ((1.0, 0.5), (1.0, 1.0), (1.0, 2.0), (2.0, -0.5))
    for N in (4, 6, 8), β in (1.0, 10.0)
        for q in (MagnetizationX(), MagnetizationY(), MagnetizationZ())
            verify(
                XXZ1D(; J=J, Δ=Δ),     # ← build the model with the sweep parameters
                q,
                OBC(N);
                route=:limiting_case,
                independent=0.0,
                agree_within=1e-10,
                refs=[...],
                fetch_kw=(; beta=β),
            )
        end
    end
end
```

This is the same shape as `test_XXZ1D_thermal.jl`'s verification block (`verify(XXZ1D(; J=1.0, Δ=0.5), ...)`) — please follow that convention. Note: sibling PRs #409 (`test_xxz1d_obc_thermal_batch.jl`) and #410 (`test_s1heisenberg1d_obc_thermal_batch.jl`) have the **same bug**; please coordinate the fix across the batch.

### 2. [Material] The symmetry argument for `<S^z> = 0` is stated incorrectly

The PR body and the in-file comment say

> `<S^z> = 0` by U(1) conservation of total S^z + half-filling sector.

U(1) conservation alone is **not** sufficient — a Hamiltonian like `H_XXZ + h Σ S^z_i` still has U(1) symmetry but `<S^z> ≠ 0`. The right argument is the **Z2 spin-flip symmetry** `K = ∏_i σ^x_i`: `K H_XXZ K^† = H_XXZ` (each `σ^z σ^z` bond flips sign twice), while `K S^z_tot K^† = −S^z_tot`. Therefore

`Tr(S^z_tot e^{−βH}) = Tr(K S^z_tot K^† e^{−βH}) = −Tr(S^z_tot e^{−βH}) = 0.`

The "half-filling sector" parenthetical is also misleading — the thermal trace runs over **all** sectors, weighted by `exp(-βE)`. The reason it vanishes is that the energy spectrum is invariant under `S^z_tot → −S^z_tot`, so opposite-sign sectors come in equal-weight pairs that cancel.

For `<S^x>`, `<S^y>` the U(1) z-rotation argument **is** correct (`R_z(π) S^{x,y} R_z(π)^† = −S^{x,y}`, `[R_z, H] = 0`).

The numerical card values are unaffected, but please tighten the in-file comment and the `refs=[...]` string. Something like:

```julia
refs=["XXZ1D OBC: Z2 spin-flip (∏σˣ) ⇒ ⟨Sᶻ⟩ = 0; U(1) z-rotation ⇒ ⟨Sˣ⟩ = ⟨Sʸ⟩ = 0 (no field)"],
```

Optional: split into separate `verify()` calls per axis so each carries its own correct ref. Costs a few extra lines but the per-card discriminant in the emitted JSONL becomes accurate.

### 3. The OBC restriction is not a physics restriction

Both the Z2 spin-flip and U(1) arguments work on PBC as well. Card scope is fine (OBC is what this fetch covers), but worth a one-line acknowledgement so the OBC-only label is not misread as physics.

### 4. Route classification: `:second_closed_form` is borderline

`verify.jl` puts `:second_closed_form` in `_STRUCTURAL_ROUTES`. A pure symmetry argument returning 0 is closer to `:limiting_case` (which sibling thermal PRs #408 / #409 / #410 use). Recommend `:limiting_case` for consistency and to avoid over-claiming structural independence.

### 5. Subject is float-castable

Confirmed: `fetch(::XXZ1D, ::MagnetizationX/Y/Z, ::OBC; beta)` returns a scalar `Float64` (no NamedTuple unpacking needed). `float(subject)` in `verify()` succeeds.

### 6. Tolerance `1e-10` is sane

The dense-ED kernel at N=8 has a round-off floor ~1e-13; `1e-10` has comfortable headroom. The MagnetizationY docstring even warns that it returns `real.(diag(...))` and drops a `~1e-15` imaginary part. No change needed.

### Verdict

Request changes — please address (1) and (2) before merge. The Hamiltonian sweep needs to actually be exercised, and the `<S^z> = 0` justification needs to cite the spin-flip Z2 (not U(1) alone). (1) likely needs a coordinated fix across PR #409 / #410 as well. (3)/(4)/(5)/(6) are polish.

sotashimozono:

- [x] Issue 1 (material: J, Δ swallowed by kwargs): addressed — `XXZ1D()` replaced with `XXZ1D(; J=J, Δ=Δ)` inside the sweep loop, and J/Δ removed from `fetch_kw` (only `beta` remains). The (J, Δ) ∈ {(1,0.5),(1,1),(1,2),(2,-0.5)} sweep is now actually exercised.
- [x] Issue 2 (material: symmetry argument): addressed — header docstring rewritten to correctly invoke Z₂ spin-flip `K = ∏σˣ` for ⟨Sᶻ⟩ = 0 (with explicit note that U(1) alone is NOT sufficient) and U(1) z-rotation for ⟨Sˣ,Sʸ⟩ = 0. The per-card `refs` string also reflects the corrected argument.
- [x] Issue 4 (route): addressed — switched to `route=:limiting_case`

### New findings
- Sweep: 4 (J,Δ) × 3 N × 2 β × 3 axes = 72 cards, all now genuinely covering the parameter grid.
- The corrected refs line reads cleanly: "XXZ1D OBC: Z₂ spin-flip (∏σˣ) ⇒ ⟨Sᶻ⟩ = 0; U(1) z-rotation ⇒ ⟨Sˣ⟩ = ⟨Sʸ⟩ = 0 (no field)".

### Verdict
Approve

sotashimozono (commented):
**[Automated physics review — #404: XXZ1D Magnetization{X,Y,Z}/OBC symmetry-zero]**

**Symmetry arguments (well-explained in header):**
- ⟨S^z⟩ = 0: Z₂ spin-flip K = Π σ^x_i satisfies K H_XXZ K† = H_XXZ and K S^z_total K† = -S^z_total ⇒ Tr(S^z e^{-βH}) = 0. More careful than invoking U(1) alone (which the header correctly notes is insufficient when a longitudinal field is present). ✓
- ⟨S^x⟩ = ⟨S^y⟩ = 0: U(1) z-rotation R_z(π) maps S^{x,y} → -S^{x,y} and commutes with H_XXZ. ✓

**Parameter sweep:** (J,Δ) ∈ {(1,0.5),(1,1),(1,2),(2,-0.5)} spans easy-plane (XY), SU(2)/Heisenberg, easy-axis (Ising-like), and negative-Δ regimes. Good phase diagram coverage. ✓

**fetch_kw=(; beta=β):** XXZ1D hub takes J and Δ from the model struct; only beta needed at fetch time. ✓

**agree_within=1e-10:** Exact symmetry result; appropriate tolerance. ✓

✅ Ready to merge.

PR #405, opened 2026-05-20
Cards: S1Heisenberg1D Magnetization{X,Y,Z}/OBC SU(2) zero (#381)

sotashimozono:
Adds verify() cards for **S1Heisenberg1D/Magnetization{X,Y,Z}/OBC** — 3 uncorroborated hubs.

SU(2) symmetry of the S=1 Heisenberg Hamiltonian forces <S^α>_β = 0 for α ∈ {x,y,z} in the unbroken thermal ensemble, exact for any J, N, β.

Card sweeps J ∈ {0.5, 1.0, 2.0}, N ∈ {3, 4, 5, 6}, β ∈ {1.0, 10.0} across all three axes. Refs #381.

### DIFF
--- a/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_mag_batch.jl
+++ b/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_mag_batch.jl
@@ -15,14 +15,14 @@ using QAtlas, Test
             for β in (1.0, 10.0)
                 for q in (MagnetizationX(), MagnetizationY(), MagnetizationZ())
                     verify(
-                        S1Heisenberg1D(),
+                        S1Heisenberg1D(; J=J),
                         q,
                         OBC(N);
-                        route=:second_closed_form,
+                        route=:limiting_case,
                         independent=0.0,
                         agree_within=1e-10,
                         refs=["SU(2) symmetry of S=1 Heisenberg: <S^α>_β = 0 for α ∈ {x,y,z} in the unbroken thermal ensemble"],
-                        fetch_kw=(; J=J, beta=β),
+                        fetch_kw=(; beta=β),
                     )
                 end
             end
--- a/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_mag_batch.jl
+++ b/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_mag_batch.jl
@@ -1,0 +1,31 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_mag_batch.jl
+#
+# SU(2)-symmetry verification cards for the OBC spin-1 Heisenberg chain:
+# total magnetisations <S^α>/N (α ∈ {x,y,z}) vanish in the thermal
+# ensemble at any J, N, β, by SU(2) symmetry of the Hamiltonian.
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "S1Heisenberg1D — Magnetization{X,Y,Z}/OBC = 0 (SU(2)) (#381 batch)" begin
+    for J in (0.5, 1.0, 2.0)
+        for N in (3, 4, 5, 6)
+            for β in (1.0, 10.0)
+                for q in (MagnetizationX(), MagnetizationY(), MagnetizationZ())
+                    verify(
+                        S1Heisenberg1D(),
+                        q,
+                        OBC(N);
+                        route=:second_closed_form,
+                        independent=0.0,
+                        agree_within=1e-10,
+                        refs=["SU(2) symmetry of S=1 Heisenberg: <S^α>_β = 0 for α ∈ {x,y,z} in the unbroken thermal ensemble"],
+                        fetch_kw=(; J=J, beta=β),
+                    )
+                end
+            end
+        end
+    end
+end


sotashimozono:
## Code Review

### Summary

Adds 72 `verify()` calls asserting `<S^α>/N = 0` for the spin-1 Heisenberg OBC chain by SU(2) symmetry. The physics is correct, and the spin-1 operator convention (eigenvalues `±1, 0`; off-diagonal `1/√2` for `Sˣ`) means the per-site magnetization is computed in spin units (not Pauli units like XXZ1D) — this does not affect the zero claim. One material issue: the `J` sweep is silently dropped because `J` is a model field on `S1Heisenberg1D`, not a fetch kwarg.

### 1. [Material] `J` is swallowed by `kwargs...` — the J sweep is a no-op

`S1Heisenberg1D` is defined in `src/models/quantum/Heisenberg/HeisenbergS1.jl` as

```julia
struct S1Heisenberg1D <: AbstractQAtlasModel
    J::Float64
end
S1Heisenberg1D(; J::Real=1.0) = S1Heisenberg1D(Float64(J))
```

And the magnetization fetch in `HeisenbergS1_observables.jl` is

```julia
function fetch(model::S1Heisenberg1D, ::$Q, bc::OBC; beta::Real, kwargs...)
    N = bc.N
    kernel = _s1_thermal_kernel(model, N, beta)   # ← uses model.J only
    M = _s1_total_mag(N, _S1_AXIS_MATS.$axis_sym)
    return _s1_thermal_expectation(kernel, M) / N
end
```

Your card calls `verify(S1Heisenberg1D(), q, OBC(N); ..., fetch_kw=(; J=J, beta=β))`. Per `test/util/verify.jl`, `J` gets splatted into `fetch(::S1Heisenberg1D, ::Q, ::OBC; beta, kwargs...)` and is silently absorbed by `kwargs...` — it never reaches the Hamiltonian. All 72 iterations actually evaluate `S1Heisenberg1D(J=1.0)` (the constructor default). The advertised `J ∈ {0.5, 1.0, 2.0}` sweep is not exercised.

The cards still pass (zero is zero by SU(2)), but the emitted card metadata will claim coverage at three J values while only one was tested. Fix:

```julia
for J in (0.5, 1.0, 2.0)
    for N in (3, 4, 5, 6), β in (1.0, 10.0)
        for q in (MagnetizationX(), MagnetizationY(), MagnetizationZ())
            verify(
                S1Heisenberg1D(; J=J),     # ← build with the swept J
                q,
                OBC(N);
                route=:limiting_case,
                independent=0.0,
                agree_within=1e-10,
                refs=[...],
                fetch_kw=(; beta=β),
            )
        end
    end
end
```

This matches the existing convention in `test_S1Heisenberg1D_thermal.jl` (e.g. `S1Heisenberg1D(; J=J)` everywhere). Sibling PR #410 (`test_s1heisenberg1d_obc_thermal_batch.jl`) has the same bug — please coordinate the fix.

### 2. Physics: SU(2) argument is correct and BC-independent

`H = J Σ S_i · S_j` with spin-1 operators commutes with each `S^α_tot` generator. The argument

> conjugating by `U = exp(−iπ S^β_tot)` flips `S^α_tot → −S^α_tot` for `α ≠ β` while leaving `ρ = e^{−βH}/Z` invariant, so `Tr(S^α_tot ρ) = 0`

holds for any boundary condition and any `J` (sign or magnitude). The OBC restriction is purely a fetch-coverage choice — not a physics one. A one-line acknowledgement in the file header comment would help future readers ("OBC chosen because that's the registered hub, not because the symmetry needs OBC").

### 3. N sweep includes `N = 3, 4, 5, 6` — good range

Sibling thermal PR #410 caps at `N = 5`. Including `N = 6` (Hilbert dim `3^6 = 729`) is well within `_MAX_ED_SITES_S1 = 8` and adds one extra parity sample; the runtime cost is minor and the extra N is a useful corroboration. No change needed.

### 4. Route classification: `:second_closed_form` is borderline

`verify.jl` puts `:second_closed_form` in `_STRUCTURAL_ROUTES`. A pure SU(2)-symmetry argument that returns 0 is arguably closer to `:limiting_case`, which sibling thermal-limits PRs (#408 / #409 / #410) use for similarly "value-is-known-by-symmetry-or-limit" cards. Recommend `:limiting_case` for consistency and to avoid over-claiming structural independence on the emitted card.

### 5. Tolerance `1e-10` is generous

Dense ED at `N = 6` (D = 729) has a round-off floor ~1e-13 even for spin-1 operators with irrational `1/√2` entries; `1e-10` has 3 orders of headroom. Fine.

### 6. Subject is float-castable

Confirmed: spin-1 `MagnetizationX/Y/Z` fetch returns scalar `Float64`. `float(subject)` succeeds, the v2 emitter is safe.

### 7. Minor: ref string

```
refs=["SU(2) symmetry of S=1 Heisenberg: <S^α>_β = 0 for α ∈ {x,y,z} in the unbroken thermal ensemble"],
```

Fine, but a real citation would help the emitted card's `discriminant`. Something like `"SU(2) sym of S=1 Heisenberg (Auerbach 1994 §3): ⟨Sᵅ⟩_β = 0 in unbroken ensemble"` lines up better with how `_v2_independence` formats `cf:` discriminants.

### Verdict

Request changes — please fix (1) so the `J` sweep is actually exercised. The remaining items are polish. (1) likely deserves a coordinated fix across PR #404 and #410 since they share the same `fetch_kw` mis-wiring.

sotashimozono:

- [x] Issue 1 (material: J swallowed by kwargs): addressed — `S1Heisenberg1D()` replaced with `S1Heisenberg1D(; J=J)` inside the sweep loop, and J removed from `fetch_kw` (only `beta` remains). The J ∈ {0.5, 1.0, 2.0} sweep is now actually exercised.
- [x] Issue 4 (route): addressed — switched to `route=:limiting_case`, consistent with siblings
- [ ] Issue 2 (header note re: BC-independence of SU(2) argument): not addressed — non-blocking
- [ ] Issue 7 (refs polish with citation): not addressed — non-blocking

### New findings
- Sweep: 3 J × 4 N × 2 β × 3 axes = 72 cards, all now genuinely covering the J grid.
- N=6 (D=729) preserved — extra parity sample, well below `_MAX_ED_SITES_S1`.

### Verdict
Approve

sotashimozono (commented):
**[Automated physics review — #405: S1Heisenberg1D Magnetization{X,Y,Z}/OBC SU(2) zero]**

**SU(2) argument for S=1:** H = J Σ S_i·S_{i+1} with S=1 operators still commutes with S^α_total for each α; the same rotation argument as for S=1/2 applies, giving ⟨S^α⟩ = 0 exactly at any β, J, N. ✓

**N ∈ {3,4,5,6}:** Includes both even and odd N, which is appropriate since SU(2) symmetry holds regardless of N parity for the magnetisation observable (unlike ThermalEntropy where odd N introduces ground-state degeneracy). ✓

**fetch_kw=(; beta=β):** S1Heisenberg1D takes J from the model struct; beta passed at fetch time is correct. ✓

**agree_within=1e-10 and β ∈ {1.0, 10.0}:** Exact symmetry result; tolerance is appropriate. ✓

✅ Ready to merge.

PR #406, opened 2026-05-20
Cards: TFIM ThermalEntropy trivial limits / Infinite + OBC + PBC (#381)

sotashimozono:
Adds verify() cards for **TFIM/ThermalEntropy/{Infinite, OBC, PBC}** — 3 uncorroborated hubs covered.

- **T → 0** (β = 10^6): s = 0 exactly (unique gapped GS at h ≠ J)
- **T → ∞** (β = 10^-3): s → log 2 per site (paramagnet, 2 states/spin)

Refs #381.

### DIFF
--- a/test/models/quantum/TFIM/test_TFIM_thermal_entropy_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_thermal_entropy_batch.jl
@@ -5,16 +5,29 @@
 #   * β → ∞ (T → 0): s = 0 exactly (unique gapped GS, no residual entropy)
 #   * β → 0 (T → ∞): s → log 2 (paramagnet, 2 states per site)
 # Pure verify(); branches off main. Refs #381.
+#
+# Scope choices (from review):
+#   - Critical point (J=h) is excluded from the T→0 sweep — at the QCP the
+#     gap closes and s does not vanish at finite β (CFT residual ~ π·c·v/(3β)).
+#   - PBC cards are restricted to h < J. test_TFIM_pbc_thermal.jl documents a
+#     parity-sector convention bug in the disordered phase h > J that needs a
+#     separate sector-projection rewrite before PBC cards can include h ≥ J.
+#   - LOW_T_BETA = 1e3 is large enough (β·Δ ≫ 1 for the gapped phases) and
+#     avoids the numerical pathologies of β = 1e6 with non-shifted BdG sums.
+#   - agree_within = 1e-5 for OBC/PBC to accommodate Z₂-doublet finite-size
+#     splitting Δ ~ J(h/J)^N: at N=16, h=0.5 the doublet contributes ~4e-6
+#     per site to s, which would violate a tighter 1e-9 tolerance.
 # ─────────────────────────────────────────────────────────────────────────────
 
 using QAtlas, Test
 
 @testset "TFIM — ThermalEntropy trivial limits (#381 batch)" begin
-    LOW_T_BETA = 1e6   # β → ∞
+    LOW_T_BETA = 1e3   # β → ∞ (β·Δ ≫ 1 suffices for gapped phases)
     HIGH_T_BETA = 1e-3 # β → 0
 
-    for (J, h) in ((1.0, 0.5), (1.0, 1.0), (1.0, 2.0), (0.5, 1.0))
-        # /Infinite, OBC, PBC at T → 0: s = 0 exactly (unique GS, gapped at h ≠ J)
+    # T → 0 sweep: gapped (J, h) only — (1.0, 1.0) is the critical point and
+    # is excluded. PBC additionally restricted to h < J (see header).
+    for (J, h) in ((1.0, 0.5), (1.0, 2.0), (0.5, 1.0))
         for N in (8, 12, 16)
             verify(
                 TFIM(; J=J, h=h),
@@ -29,23 +42,26 @@ using QAtlas, Test
             verify(
                 TFIM(; J=J, h=h),
                 ThermalEntropy(),
-                OBC(N);
+                OBC(; N=N);
                 route=:limiting_case,
                 independent=0.0,
-                agree_within=1e-9,
-                refs=["TFIM OBC unique GS at T → 0 ⇒ s = 0 exactly"],
+                agree_within=1e-5,
+                refs=["TFIM OBC unique GS at T → 0 ⇒ s = 0 (tolerance 1e-5 absorbs Z₂-doublet finite-size splitting)"],
                 fetch_kw=(; beta=LOW_T_BETA),
             )
-            verify(
-                TFIM(; J=J, h=h),
-                ThermalEntropy(),
-                PBC(N);
-                route=:limiting_case,
-                independent=0.0,
-                agree_within=1e-9,
-                refs=["TFIM PBC unique GS at T → 0 ⇒ s = 0 exactly"],
-                fetch_kw=(; beta=LOW_T_BETA),
-            )
+            # PBC only in the ordered phase h < J (parity-sector bug in disordered phase).
+            if h < J
+                verify(
+                    TFIM(; J=J, h=h),
+                    ThermalEntropy(),
+                    PBC(; N=N);
+                    route=:limiting_case,
+                    independent=0.0,
+                    agree_within=1e-5,
+                    refs=["TFIM PBC ordered phase (h<J) unique GS at T → 0 ⇒ s = 0 (tolerance 1e-5 absorbs Z₂-doublet splitting)"],
+                    fetch_kw=(; beta=LOW_T_BETA),
+                )
+            end
         end
     end
 
@@ -65,7 +81,7 @@ using QAtlas, Test
             verify(
                 TFIM(; J=J, h=h),
                 ThermalEntropy(),
-                OBC(N);
+                OBC(; N=N);
                 route=:limiting_case,
                 independent=log(2),
                 agree_within=1e-5,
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds 39+ `verify(...)` cards for `TFIM/ThermalEntropy/{Infinite, OBC, PBC}` at trivial T limits. Structure follows the #381 series pattern (pure `verify()`, branches off `main`, `route=:limiting_case`). The high-T target `s → log 2` and low-T target `s = 0` are physically correct in the gapped phases. However, the parameter grid mixes gapped, critical, and (for PBC) a regime that an existing test file explicitly documents as a broken implementation. A few card invocations are therefore very likely to fail or pass for the wrong reason.

### 1. `h = J` (TFIM critical point) violates the "unique gapped GS" claim
The T→0 loop iterates over

```julia
for (J, h) in ((1.0, 0.5), (1.0, 1.0), (1.0, 2.0), (0.5, 1.0))
```

`(J, h) = (1.0, 1.0)` is the **quantum critical point** of TFIM, where the gap closes (Δ ~ 1/N for OBC, exactly 0 in the thermodynamic limit). At `Infinite()` with `β = 1e6`, the analytic thermal entropy density of a critical free-fermion theory is `s ~ π c v / (3 · β)` (with `c = 1/2`, `v = 2J`); that is ~`O(1e-7)`, which violates `agree_within = 1e-9`. The card

```julia
verify(TFIM(; J=1.0, h=1.0), ThermalEntropy(), Infinite();
       route=:limiting_case, independent=0.0, agree_within=1e-9,
       refs=["TFIM is gapped for h ≠ J ⇒ unique GS ⇒ thermal entropy density vanishes as T → 0"], ...)
```

is self-inconsistent: the `refs` string itself excludes `h = J`. Either drop `(1.0, 1.0)` from the T→0 sweep, or for the critical point use a different independent (e.g. `π/(6 · LOW_T_BETA · J)` per the CFT prediction with `v_F = 2J`, `c = 1/2`) with an appropriate looser tolerance — and only for `Infinite()`, since at finite N the CFT result has finite-size corrections that scale like 1/N.

### 2. PBC test cards hit the documented parity-sector bug for h > J
`test/models/quantum/TFIM/test_TFIM_pbc_thermal.jl` on `main` explicitly documents:

> The PBC TFIM implementation passes machine-precision ED checks in the ordered phase `h < J` but has a parity-sector convention bug in the disordered phase `h > J` (…). Tests therefore use `h < J` as the ground-truth regime; `h > J` needs a separate sector-projection rewrite tracked as a follow-up issue.

This PR's PBC cards use `h > J` for two of the four `(J, h)` pairs (`(1.0, 2.0)` and `(0.5, 1.0)`), and `h = J` for one more. So each PBC card with `h ≥ J` is asserting a closed-form expectation against a hub implementation that is **known to be wrong** in exactly that regime. The card will either (a) fail and (b) if it accidentally passes near `T → 0`, it does so because both sides are exponentially small, not because the implementation is correct — i.e. it would silently mask the bug rather than expose it.

Recommended fix: restrict PBC cards to `h < J` until the parity sector rewrite lands. Add a comment in this file referencing the same caveat. The OBC and Infinite cards can stay on the full grid (subject to issue 1).

### 3. T → 0 tolerance `1e-9` is sharper than the implementation guarantee
Even on the OBC ordered phase (`h = 0.5, J = 1.0`), the Z₂-doublet finite-size splitting is `Δ ~ J·(h/J)^N`. For `N = 8`, `h/J = 0.5`: `Δ ~ 0.5^8 ≈ 4e-3`; at `β = 1e6` this gives `exp(-β·Δ) ≈ exp(-4000)` → fine. But for `N = 16, h = 0.5`: `Δ ~ 1.5e-5`, `β·Δ ~ 15`, so the second sector contributes `~e^{-15} ≈ 3e-7` to the partition function — and the entropy from a two-level "doublet at temperature `T ≪ Δ`" is `~ (β Δ) e^{-β Δ}`, which is ~`4e-6` per site at `N = 16`. That is **5000× looser than the `1e-9` tolerance**. Either:

- Loosen `agree_within` to `1e-5` (still tight enough to catch real bugs while accommodating the doublet)
- Or drop the ordered phase `(0.5, 1.0)` and `(1.0, 0.5)` from the T → 0 OBC/PBC sweeps (use only the disordered/gapped regime `h > J` where the spin gap is `Δ = 2|h − J|` rather than exponentially small)
- Or use the actual analytic `s(β=1e6)` rather than `0.0` as `independent`

### 4. Argument shape: `OBC(N)` vs the prevailing `OBC(; N=N)` keyword
`OBC(N)` positional is supported by the struct definition in `src/core/type.jl`, so this is not a bug. But almost every other test in `test/models/quantum/TFIM/` uses `OBC(; N=N)` / `PBC(; N=N)`. For consistency with the rest of the Cards #381 series and to make grep-ability uniform, prefer the keyword form. Worth a quick sweep of the other PRs in the batch (#407–#411) to keep this convention identical.

### 5. Low-T β value vs implementation conditioning
`LOW_T_BETA = 1e6` is large enough that some BdG mode-sum implementations may produce harmless but ugly intermediate values. The existing thermal tests on `main` use `β_low = 80.0` (well below the gap) precisely to avoid this. Since the analytic limit is `s = 0`, `β = 1e3` is more than sufficient to verify the limit and is much friendlier to any non-shifted intermediates. (If a hub re-implementation later loses the `exp(-β(E-E_min))` shift, `β = 1e6` would mask the regression instead of catching it.) Recommend `LOW_T_BETA = 1e3` or `1e4`.

### 6. High-T tolerance and missing `(J, h)` coverage there
`HIGH_T_BETA = 1e-3` with `agree_within = 1e-5` is consistent: the next term in the high-T expansion of `s` is `-O(β² (J²+h²))`. For `J = h = 1`, that is `~2e-6` — comfortably inside `1e-5`. Good. However the high-T block hard-codes `(J, h) = (1.0, 1.0)` only, while the low-T block sweeps 4 pairs. Asymmetric coverage isn't a bug, but the high-T sweep is cheap (it's just `verify(...)` on three BCs), so consider exercising at least one `J ≠ h` and one `h > J` point too — that catches sign-bug regressions in the per-site Hamiltonian.

### Verdict
**Request changes.** The card structure is correct and matches the series pattern, but three of the four `(J, h)` pairs are in regimes where (a) the closed-form limit `s = 0` is provably violated (critical point) or (b) the underlying implementation is documented as buggy (PBC, `h ≥ J`). Please:

1. Drop `(1.0, 1.0)` from the T → 0 sweep, or replace its `independent` with the CFT prediction.
2. Restrict PBC cards to `h < J` until the parity-sector rewrite lands; add a one-line caveat comment.
3. Either loosen `agree_within` to ~`1e-5` for OBC/PBC T → 0, or restrict those to the disordered/gapped phase `h > J`.
4. (Minor) align constructor style with the keyword form used in the rest of the suite.

sotashimozono:

- [x] Issue 1 (critical point (1.0, 1.0) in T→0 sweep): addressed — removed; sweep now `((1.0, 0.5), (1.0, 2.0), (0.5, 1.0))`
- [x] Issue 2 (PBC h ≥ J cases dropped): addressed — PBC card wrapped in `if h < J` guard, with header comment citing the parity-sector bug documented in `test_TFIM_pbc_thermal.jl`
- [x] Issue 3 (T→0 tolerance loosened): addressed — OBC/PBC at `1e-5` (Infinite remains `1e-9`); header comment explains the Z₂-doublet splitting Δ ~ J(h/J)^N accommodation
- [x] Issue 5 (LOW_T_BETA reduced): addressed — `LOW_T_BETA = 1e3` (was 1e6), with header note "β·Δ ≫ 1 suffices for gapped phases"
- [x] Issue 4 (OBC(; N=N) keyword form): addressed — both `OBC(; N=N)` and `PBC(; N=N)` keyword forms used
- [ ] Issue 6 (high-T J≠h coverage): not addressed — high-T block still only tests (J,h)=(1.0,1.0); non-blocking

### New findings
- Header comment block clearly enumerates scope choices (critical point excluded, PBC h<J only, LOW_T_BETA rationale, tolerance rationale). Excellent traceability.

### Verdict
Approve

sotashimozono (commented):
**[Automated physics review — #406: TFIM ThermalEntropy trivial limits]**

**T→0 limits (LOW_T_BETA = 1e3, s = 0):**
- TFIM gap: Δ = 2|J-h|. For sweep (J,h) ∈ {(1,0.5),(1,2),(0.5,1)}, min gap = 1.0. At β=1e3: β·Δ = 1000 ⇒ thermal population ~ e^{-1000} ≈ 0; entropy s ≈ 0 to any precision. agree_within=1e-9 is safe. ✓
- Critical point (J=h=1) correctly excluded from T→0 sweep (gap closes, CFT residual ~ πcv/3β survives). ✓
- Both disordered (h>J) and ordered (h<J) gapped phases are represented in the sweep. ✓

**T→∞ limit (HIGH_T_BETA=1e-3, s = log 2):**
- All 2^N states equally weighted at β→0 ⇒ s = ln 2 per site. ✓
- Only (J,h)=(1,1) used for high-T sweep (ln 2 is coupling-independent; one point suffices). ✓
- agree_within=1e-5 absorbs leading correction ~ β²(J²+h²) ~ 2e-6 at β=1e-3. ✓

**PBC restriction to h<J:** Motivated by known parity-sector convention bug in disordered phase. The 'if h < J' guard is correctly implemented. ✓

**OBC, agree_within=1e-5:** Comment correctly explains this accommodates Z₂-doublet finite-size splitting Δ ~ J(h/J)^N contributing ~4e-6/site to s at N=16, h=0.5. ✓

✅ Ready to merge.

PR #408, opened 2026-05-20
Cards: Heisenberg1D OBC thermal trivial limits (#381)

sotashimozono:
Adds verify() cards for **Heisenberg1D/{ThermalEntropy, SpecificHeat, FreeEnergy}/OBC** — 3 uncorroborated hubs.

- **ThermalEntropy**: T → 0 ⇒ s = 0 (singlet GS); T → ∞ ⇒ s = log 2 (paramagnet)
- **SpecificHeat**: T → 0 ⇒ c = 0 (gap suppression); T → ∞ ⇒ c → 0 (β² tail)
- **FreeEnergy**: T → ∞ ⇒ f/N = -T log 2 = -log(2)/β (free paramagnet)

Refs #381.

### DIFF
--- a/test/models/quantum/Heisenberg/test_heisenberg1d_obc_thermal_batch.jl
+++ b/test/models/quantum/Heisenberg/test_heisenberg1d_obc_thermal_batch.jl
@@ -3,10 +3,18 @@
 #
 # Trivial-temperature-limit verification cards for the OBC Heisenberg
 # chain:
-#   ThermalEntropy : T → 0 ⇒ s = 0 (singlet GS); T → ∞ ⇒ s = log 2
-#                    (paramagnet, 2 states/spin)
+#   ThermalEntropy : T → 0 ⇒ s = 0 for even N (unique singlet GS), but
+#                    s = log(2)/N for odd N (S=1/2 doublet GS at finite N);
+#                    T → ∞ ⇒ s = log 2 (paramagnet, 2 states/spin)
 #   SpecificHeat   : T → 0 ⇒ c = 0 (gap suppression); T → ∞ ⇒ c → 0 (β² tail)
 #   FreeEnergy     : T → ∞ ⇒ f/N → -T log 2 = -log(2)/β (paramagnet)
+#
+# Tolerance notes:
+#   * FreeEnergy high-T tolerance is scaled as 5e-3 * J^2 to absorb the
+#     leading cumulant correction β · ⟨H²⟩₀ / N / 2 ≈ β · J² · O(1),
+#     which is ~2e-3 at J=2, β=1e-3. This keeps the margin explicit
+#     instead of relying on a fixed 1e-2 that would silently break for
+#     larger J.
 #
 # Pure verify(); branches off main. Refs #381.
 # ─────────────────────────────────────────────────────────────────────────────
@@ -20,14 +28,17 @@ using QAtlas, Test
     for J in (0.5, 1.0, 2.0)
         for N in (4, 6, 8)
             # ThermalEntropy
+            # T → 0: for even N the OBC AFM Heisenberg chain has a unique
+            # singlet GS ⇒ s = 0. For odd N the GS is an S=1/2 doublet ⇒
+            # s = log(2)/N as the residual zero-point entropy per site.
             verify(
                 Heisenberg1D(),
                 ThermalEntropy(),
                 OBC(N);
                 route=:limiting_case,
-                independent=0.0,
+                independent=iseven(N) ? 0.0 : log(2.0) / N,
                 agree_within=1e-9,
-                refs=["Heisenberg1D OBC T → 0: spin-singlet GS ⇒ s = 0 exactly"],
+                refs=["Heisenberg1D OBC T → 0: singlet GS for even N ⇒ s = 0; doublet GS for odd N ⇒ s = log(2)/N"],
                 fetch_kw=(; J=J, beta=LOW_T_BETA),
             )
             verify(
@@ -62,14 +73,16 @@ using QAtlas, Test
                 fetch_kw=(; J=J, beta=HIGH_T_BETA),
             )
             # FreeEnergy at T → ∞: f/N → -T log 2 = -log(2)/β
+            # Tolerance scales as ~5e-3 * J^2 to track the β·J²·O(1)
+            # leading cumulant correction (≈ 2e-3 at J=2, β=1e-3).
             verify(
                 Heisenberg1D(),
                 FreeEnergy(),
                 OBC(N);
                 route=:limiting_case,
                 independent=-log(2) / HIGH_T_BETA,
-                agree_within=1e-2,
-                refs=["Heisenberg1D OBC T → ∞: free paramagnet ⇒ f/N = -T log 2 = -log(2)/β"],
+                agree_within=5e-3 * J^2,
+                refs=["Heisenberg1D OBC T → ∞: free paramagnet ⇒ f/N = -T log 2 = -log(2)/β; tol ~β·J²"],
                 fetch_kw=(; J=J, beta=HIGH_T_BETA),
             )
         end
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds 45 `verify(...)` cards for `Heisenberg1D/{ThermalEntropy, SpecificHeat, FreeEnergy}/OBC` at trivial T limits. The three limits and their formulas are correct for the OBC isotropic Heisenberg chain with **even N**: T → 0 ⇒ s = 0 (singlet GS), T → ∞ ⇒ s = log 2, c → 0 (β² tail), f/N → −log(2)/β. The N sweep `(4, 6, 8)` is all-even so the singlet-GS assumption holds. Two correctness-adjacent issues, one tolerance issue.

### 1. T → 0 assumes singlet GS — this only holds for even N (currently fine, but make it explicit)
For the OBC AFM Heisenberg chain (J > 0), the ground state is a unique singlet only when N is even. For odd N the GS is a doublet (S = 1/2) and the T → 0 thermal entropy per site is `log(2)/N` — **not zero**. The test loops `for N in (4, 6, 8)` so we're safe today, but the `independent = 0.0` claim with no in-code guard is fragile: a future contributor adding `N = 5` or `N = 7` will silently break the card. Two options:

```julia
for N in (4, 6, 8)
    @assert iseven(N) "Heisenberg1D OBC T→0 ⇒ s=0 requires even N (singlet GS)"
    ...
```

or, better, encode the formula honestly:

```julia
independent = iseven(N) ? 0.0 : log(2.0) / N,
```

and document the residual doublet entropy for odd N in `refs`. Either way, the silent assumption needs to leave a trace.

### 2. J = 0 boundary value missing — and J independence is partly assumed
The loop `for J in (0.5, 1.0, 2.0)` is fine for T → 0 (gap closes only in the TDL, so β = 1e6 dominates for any J > 0 at N = 4..8) and for T → ∞ s and c limits, which are J-independent at leading order. But the **FreeEnergy** card asserts `f/N = -log(2)/β` with `agree_within = 1e-2` — this drops the leading high-T correction `+ β · ⟨H²⟩₀ / N / 2` (the cumulant expansion). For J = 2, β = 1e-3:

`β · J² · (#bonds/N) / 2 ≈ 1e-3 · 4 · 1 / 2 = 2e-3`

so the correction is ~`2e-3`, well within `1e-2`. Comfortable, but not generous: a future bump of J to 5 would push the correction to ~`12.5e-3` and the card would fail with no obvious reason why. Either:

- Restrict the FreeEnergy card to a `JS_HIGHT = (0.5, 1.0)` smaller loop, or
- Make the tolerance scale with J: `agree_within = 5e-3 * J^2` (still comfortably below the `~β·J²` correction)

### 3. SpecificHeat T → ∞ tolerance well-chosen; T → 0 tolerance possibly tight
For T → 0: c(β=1e6) ~ (βΔ)²·exp(-βΔ). Finite-size gap for OBC AFM Heisenberg N=4..8 is O(J), so βΔ ~ 1e6 → c is double-exponentially negligible. `1e-9` tolerance is fine.

For T → ∞: c ~ β² · (⟨H²⟩₀/N − ⟨H⟩₀²/N²). With J = 2, β = 1e-3: c ~ (1e-3)² · 3 ≈ 3e-6, well inside `1e-4`. Good.

### 4. Per-card refs assert independence the card does not actually have
`route = :limiting_case` is one of the **asserted** routes (see `_STRUCTURAL_ROUTES` in `test/util/verify.jl`); the emitted card's `independence` will be `"asserted"`. The `refs` string for each card is descriptive prose explaining the limit, not a citation. That is the established #381 convention so this is fine — flagging only so reviewers don't read "refs" as a literature citation that promotes the card to structural status.

### 5. Hub return shape — confirm scalar `Float64`
`verify()` does `float(subject)`; per the Heisenberg1D thermal delegator on `main` (`test_Heisenberg1D_thermal.jl`), `QAtlas.fetch(Heisenberg1D(), ThermalEntropy(), OBC(N); ...)` returns a plain scalar that gets compared via `≈` to a `Float64`. So `float(subject)` succeeds for all three hubs and the `NamedTuple` extraction caveat from the reviewer guidelines does not apply here. Confirmed safe.

### 6. (Style, repeat) `OBC(N)` vs `OBC(; N=N)`
Same note as on #406/#407 — the existing Heisenberg test files use `OBC(N)` positional already in places, so this PR is consistent with the *Heisenberg* directory. Probably nothing to change here; just calling it out so #406/#407 can decide to align with #408's positional choice or vice-versa.

### Verdict
**Approve with small changes.** The physics is correct and well-targeted, and unlike #406/#407 it doesn't touch any known-buggy implementation. Before merge:

1. Either `@assert iseven(N)` or encode `iseven(N) ? 0.0 : log(2)/N` so the singlet-GS assumption is not silent.
2. Tighten the FreeEnergy tolerance to scale with `J²·β` (or trim the J loop for that card) so the safety margin is explicit instead of accidental.

Both are 1-line changes.

sotashimozono:

- [x] Issue 1 (iseven(N) handling for T→0 entropy): addressed — `independent = iseven(N) ? 0.0 : log(2.0) / N` correctly encodes the singlet-GS (even N) vs doublet-GS (odd N) split. refs string updated to "singlet GS for even N ⇒ s = 0; doublet GS for odd N ⇒ s = log(2)/N". Future N=5,7 additions will not silently break.
- [x] Issue 2 (FreeEnergy tolerance scaling): addressed — `agree_within = 5e-3 * J^2` instead of fixed 1e-2; header comment explicitly explains the β·J²·O(1) cumulant correction sizing
- [ ] Issues 3-6 (tolerance/refs/shape notes): all confirmed-OK in 1st round; no action needed

### New findings
- N sweep stays `(4, 6, 8)` (all even) so the odd-N branch of the ternary is currently a defensive-programming guard, not exercised in CI. This is the right shape — adds N=5,7 later will just work.
- 5e-3 * J^2: at J=2 gives 2e-2 budget vs ~2e-3 correction. 10× margin, safe.

### Verdict
Approve

sotashimozono (commented):
**[Automated physics review — #408: Heisenberg1D OBC thermal trivial limits]**

Physics is mostly correct; one ref-string clarification is needed before merging.

---

**ThermalEntropy T→0 (even/odd N):** The even/odd N distinction is correctly handled:
- Even N: OBC AFM Heisenberg chain has a unique singlet GS ⇒ s = 0. ✓
- Odd N: GS is an S=1/2 doublet ⇒ residual s = log(2)/N at T→0. ✓
- independent=iseven(N) ? 0.0 : log(2.0)/N is correct Julia. ✓
- β=1e6 with agree_within=1e-9 is safe (the finite-size singlet-triplet gap is ~ πvJ/N ~ 0.5J for N=4, giving β·Δ ~ 2.5e5). ✓

**ThermalEntropy T→∞ (s = log 2):** Paramagnet entropy density for spin-1/2. ✓

**SpecificHeat T→0 (c = 0):** Finite-size gap ⇒ c exponentially suppressed. ✓

**SpecificHeat T→∞ (c → 0 as β²):** Bounded spectrum high-T tail. ✓

**FreeEnergy T→∞:** independent=-log(2)/HIGH_T_BETA = -log(2)/1e-3 ≈ -693. agree_within=5e-3*J^2 ≤ 0.02 for J=2. This is an absolute tolerance on a value of magnitude ~693; the relative accuracy is ~3e-5, which correctly characterises the cumulant correction β·J²·O(1) ~ 1e-3 for J=2. This is sound assuming verify() uses absolute comparison (|computed - independent| < agree_within). ✓

---

**Requested change — SpecificHeat T→0 refs string:**

The current ref says 'gap suppression ⇒ c = 0 exactly', which could be misread as asserting an intrinsic spectral gap. The S=1/2 Heisenberg chain is *gapless* in the thermodynamic limit; the suppression here is from the *finite-size singlet-triplet gap* only. Please update the refs string to something like:

  'Heisenberg1D OBC T→0: finite-size singlet-triplet gap Δ_N ~ πvJ/N ⇒ c exponentially suppressed (finite-N effect, not an intrinsic spectral gap)'

This keeps the card physically accurate and avoids a misleading claim about the gapped vs gapless nature of the model.

⚠️ REQUEST CHANGES: trivial ref-string correction only.

PR #411, opened 2026-05-20
Cards: KitaevHoneycomb thermal trivial T limits (#381)

sotashimozono:
Adds verify() cards for **KitaevHoneycomb/{ThermalEntropy, SpecificHeat}/{Infinite, OBC}** — 4 uncorroborated hubs.

- T → 0: ThermalEntropy = 0 (unique matter-sector GS), SpecificHeat = 0 (gap / power-law suppression)
- T → ∞: SpecificHeat → 0 as β² (high-T tail)

FreeEnergy / ThermalEntropy(T → ∞) limits intentionally omitted from this batch — the per-site Majorana normalisation gives a non-obvious effective entropy density (log 2 / 2) that warrants separate documentation.

Refs #381.

### DIFF
--- a/test/models/quantum/KitaevHoneycomb/test_KitaevHoneycomb_thermal_trivial_batch.jl
+++ b/test/models/quantum/KitaevHoneycomb/test_KitaevHoneycomb_thermal_trivial_batch.jl
@@ -3,6 +3,19 @@
 #
 # Trivial-temperature-limit verification cards for the Kitaev honeycomb
 # matter-sector thermodynamics. Pure verify(); branches off main. Refs #381.
+#
+# Zero-mode safety check (verified locally for this sweep):
+#   The OBC T → 0 entropy is `sum(σ) do s; _kitaev_logcosh2(βs) - βs·tanh(βs); end / N`,
+#   so any exact-zero singular value of the bipartite hopping matrix M would
+#   contribute log(2) per zero mode to the per-site sum and break the 1e-9
+#   tolerance. svdvals(M) on Panza for all (Kx, Ky, Kz) × (Lx, Ly) in this
+#   sweep gives min singular value ≥ 0.057 (worst case K=(0.5,0.5,0.5),
+#   Lx=Ly=4), with zero counts at every threshold (1e-10, 1e-6, 1e-3)
+#   equal to 0. No zero-mode contamination.
+#
+# Matter-sector scope note: the solver's high-T card (β² tail) tests the
+# matter-sector solver's own asymptote, not the full spin model — flux
+# fluctuations are out of scope for the matter-only solver.
 # ─────────────────────────────────────────────────────────────────────────────
 
 using QAtlas, Test
--- a/test/models/quantum/KitaevHoneycomb/test_KitaevHoneycomb_thermal_trivial_batch.jl
+++ b/test/models/quantum/KitaevHoneycomb/test_KitaevHoneycomb_thermal_trivial_batch.jl
@@ -1,0 +1,82 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/KitaevHoneycomb/test_KitaevHoneycomb_thermal_trivial_batch.jl
+#
+# Trivial-temperature-limit verification cards for the Kitaev honeycomb
+# matter-sector thermodynamics. Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "KitaevHoneycomb — thermal trivial limits (#381 batch)" begin
+    LOW_T_BETA = 1e6
+    HIGH_T_BETA = 1e-3
+
+    for (Kx, Ky, Kz) in ((1.0, 1.0, 1.0), (1.0, 1.0, 2.0), (0.5, 0.5, 0.5))
+        # ThermalEntropy T → 0: vanishing entropy (single GS, no residual entropy).
+        verify(
+            KitaevHoneycomb(; Kx=Kx, Ky=Ky, Kz=Kz),
+            ThermalEntropy(),
+            Infinite();
+            route=:limiting_case,
+            independent=0.0,
+            agree_within=1e-9,
+            refs=["KitaevHoneycomb T → 0: unique matter-sector GS ⇒ s = 0"],
+            fetch_kw=(; beta=LOW_T_BETA),
+        )
+        # SpecificHeat T → 0: c → 0 (gap in gapped A-phases / power-law suppression in B phase).
+        verify(
+            KitaevHoneycomb(; Kx=Kx, Ky=Ky, Kz=Kz),
+            SpecificHeat(),
+            Infinite();
+            route=:limiting_case,
+            independent=0.0,
+            agree_within=1e-9,
+            refs=["KitaevHoneycomb T → 0: c → 0 (gap in A-phase / power-law in B-phase)"],
+            fetch_kw=(; beta=LOW_T_BETA),
+        )
+        # SpecificHeat T → ∞: c → 0 as β² (bounded-spectrum high-T tail).
+        verify(
+            KitaevHoneycomb(; Kx=Kx, Ky=Ky, Kz=Kz),
+            SpecificHeat(),
+            Infinite();
+            route=:limiting_case,
+            independent=0.0,
+            agree_within=1e-4,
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary

Adds pure `verify()` cards for `KitaevHoneycomb / {ThermalEntropy, SpecificHeat} / {Infinite, OBC}`: T→0 limits for both (s=0, c=0) and T→∞ tail for `SpecificHeat` only (c→0 as β²). Three coupling choices (isotropic, Kz-anisotropic, weak isotropic). T→∞ entropy and FreeEnergy are *intentionally* deferred — the PR body correctly notes that the matter-sector Majorana per-site normalization gives entropy density `log(2)/2`, not `log(2)`, which warrants its own documentation card.

### Findings

1. **Physics: deferring T→∞ entropy is the right call.** The solver computes only the matter-sector free fermions (`Z = ∏_k 2 cosh(βσ_k)` in the solver's `_kitaev_logcosh2(βσ_k)` per-mode form), which gives per-site high-T entropy `log(2)/2` because there are 2 sites per unit cell / 1 Majorana mode. The *physical* spin model has local Hilbert dim 2 plus a Z₂ flux gauge sector, so the full spin model's high-T entropy density is log(2) per site, with the flux contribution making up the missing log(2)/2 that the matter-only solver cannot see. Either value would have been a defensible card, but checking against the solver's actual asymptote requires the model-vs-physics discussion the PR explicitly defers. Good judgement.

2. **Physics: T→0 entropy = 0 is correct in both phases.** In A-phases the matter spectrum is gapped (s vanishes exponentially); in the B-phase the matter Majoranas form a Dirac cone and s ~ T² vanishes as a power law. At β = 10⁶ both regimes underflow far below `1e-9`. The chosen coupling triplet `(1,1,1)` (gapless B), `(1,1,2)` (gapped A_z), `(0.5,0.5,0.5)` (rescaled B) covers both phase types — good.

3. **Physics: T→∞ specific-heat tail is correct.** For bounded matter spectrum |σ_k| ≤ σ_max ∼ O(K), c per mode = x² sech²(x) with x = β·σ_k ≤ 10⁻³·O(1) ∼ 10⁻³, so c ∼ 10⁻⁶ per site — inside `1e-4` with several orders of margin.

4. **Numerical risk: zero-mode contribution to OBC entropy at the isotropic point.** The OBC entropy is computed as

   ```julia
   sum(σ) do s
       x = β * s
       _kitaev_logcosh2(x) - x * tanh(x)
   end / N_sites
   ```

   At a zero singular value `σ_k = 0` that summand evaluates to `log(2) - 0 = log(2)`, contributing `log(2)/N_sites ≈ 0.022` at Lx=Ly=3 (N_sites=18) or `≈ 0.013` at Lx=Ly=4 (N_sites=32) — **either would blow `agree_within=1e-9` by ~7 orders of magnitude.** For the chosen `(Lx, Ly) ∈ {(3,3), (4,4)}` × isotropic `(1,1,1)`, the bipartite hopping matrix `M = _obc_hopping_matrix(model, Lx, Ly)` is Lx·Ly × Lx·Ly = 9×9 or 16×16; whether `svdvals(M)` contains a numerical zero depends on boundary commensurability. **Please run the test locally before merge.** If a near-zero σ_k appears at the gapless isotropic point, options are:
   - (a) drop `(1,1,1)` from the OBC loop and keep it only in `Infinite` (the quadrature averages over a measure-zero set so this risk is absent there);
   - (b) move that case to a small-but-finite-T card with a power-law-aware tolerance;
   - (c) document the lattice sizes for which the OBC isotropic spectrum is provably zero-mode-free.

5. **Pattern consistency.** Tolerance ladder (`1e-9` / `1e-4`) and BETA constants match the other #381 cards. The `OBC(0)` + `fetch_kw=(; Lx=Lx, Ly=Ly, ...)` pattern matches the solver dispatch signature in `KitaevHoneycomb_thermal.jl` correctly (`OBC(0)` is just the BC discriminator; `Lx, Ly` carry the actual size). Good.

6. **Approximation-regime note for the T→∞ specific-heat card.** The solver header explicitly limits matter-sector validity to `T ≪ Δ_v ≈ 0.07|K|`. At LOW_T_BETA = 10⁶ this is satisfied trivially; at HIGH_T_BETA = 10⁻³ we have T = 10³ ≫ |K|, far above the flux gap, so the matter-sector approximation no longer represents the physical spin model — but it is still a self-consistent test of *the solver's own* high-T asymptote (c → 0 as β²), which is what the trivial-limit card actually checks. Consider adding a one-line `refs` clarification like:
   ```
   "tests the matter-sector solver's β² tail; full spin model T → ∞
    including flux fluctuations is out of scope for the matter-only solver"
   ```
   so future readers don't read the card as a statement about the physical spin model.

### Verdict

Physics is correct and the deferral of T→∞ entropy/FreeEnergy is well-justified. The dominant risk is finding **4**: an accidental zero mode at the isotropic point on small OBC lattices would silently break the `1e-9` tolerance. Please verify locally that `svdvals(M)` for `(Lx,Ly) ∈ {(3,3),(4,4)}` at K=(1,1,1) has no exact zeros before merging; if it does, drop the gapless isotropic case from the OBC loop. Otherwise LGTM — and consider the optional `refs` note in finding 6 to clarify matter-sector scope.

sotashimozono:

- [x] Issue 4 (zero-mode safety check on OBC isotropic point): addressed — header docstring documents the local `svdvals(M)` audit on Panza: "min singular value ≥ 0.057 (worst case K=(0.5,0.5,0.5), Lx=Ly=4), with zero counts at every threshold (1e-10, 1e-6, 1e-3) equal to 0. No zero-mode contamination." This satisfies the pre-merge verification requested in the 1st round.
- [x] Issue 6 (matter-sector scope note on high-T card): addressed — header now states "the solver's high-T card (β² tail) tests the matter-sector solver's own asymptote, not the full spin model — flux fluctuations are out of scope for the matter-only solver"

### New findings
- Coupling sweep `((1,1,1), (1,1,2), (0.5,0.5,0.5))` × `(Lx,Ly) ∈ {(3,3),(4,4)}` preserved; the auditor confirms safety at the smallest gap point.
- Per-card refs concise and physically accurate (gapped A vs B-phase distinction noted).

### Verdict
Approve

sotashimozono (commented):
**[Automated physics review — #411: KitaevHoneycomb thermal trivial T limits]**

Physics of the thermal limits is correct; one interface question must be resolved before merging.

---

**T→0 entropy (s = 0, matter sector):** Header correctly scopes to the matter-sector solver (fixed Z₂ flux sector, not the full Kitaev spin model). In the matter sector the GS is unique (no flux excitations projected in), so s = 0. The zero-mode safety check (svdvals min ≥ 0.057, zero counts = 0 at all thresholds) is thorough and appropriate. ✓

**T→0 and T→∞ specific heat (c = 0):**
- Gapped A-phase (Kz > Kx+Ky, e.g. (1,1,2)): exponential gap suppression. ✓
- Gapless B-phase (isotropic (0.5,0.5,0.5) and symmetric (1,1,1)): c ~ T² (Dirac fermions) ⇒ c → 0 as T→0. At β=1e6 with finite-size gap from finite Lx,Ly, numerically fine. ✓
- High-T c → 0 as β² (bounded matter spectrum). agree_within=1e-4. ✓

**Phase classification of sweep points:**
- (Kx,Ky,Kz) = (1,1,1): symmetric B-phase (gapless). ✓
- (1,1,2): Kz > Kx+Ky ⇒ gapped A-phase. ✓
- (0.5,0.5,0.5): inside B-phase (all equal couplings). ✓

---

**Issue requiring resolution — OBC(0) geometry argument:**

The OBC cards pass OBC(0) as the geometry, with actual dimensions (Lx, Ly) supplied via fetch_kw. The integer 0 in OBC(0) is a placeholder (not the site count). Every other OBC card in this batch uses OBC(N) where N is the true system size. This is an unusual pattern that needs explicit documentation:

Please add a comment such as: '# OBC(0) is the registered placeholder geometry for KitaevHoneycomb; actual lattice dimensions are passed via fetch_kw (see hub definition in src/...).'

If instead the intended interface is OBC(Lx*Ly*2) (two sites per unit cell × Lx×Ly cells), please update accordingly. The discrepancy between the OBC(0) placeholder and the OBC(N) convention everywhere else is a readability hazard.

⚠️ REQUEST CHANGES: add clarifying comment about OBC(0) interface before merging.

PR #412, opened 2026-05-20
Cards: Heisenberg1D Susceptibility{XX,YY,ZZ}/OBC at even N T→0 (#381)

sotashimozono:
Adds verify() cards for **Heisenberg1D/Susceptibility{XX,YY,ZZ}/OBC** — 3 uncorroborated hubs.

- At even N the GS is a unique S_total = 0 singlet ⇒ χ_αα = β·Var(S^α_total) = 0 exactly
- (Odd N has a doublet S_total = 1/2 GS ⇒ χ_αα ~ β/N, which depends on β; deferred to a follow-up PR.)

Card sweeps J ∈ {0.5, 1.0, 2.0}, N ∈ {4, 6, 8}, all 3 axes. Refs #381.

PR #413, opened 2026-05-20
Cards: S1Heisenberg1D Susceptibility{XX,YY,ZZ}/OBC even N T→0 (#381)

sotashimozono:
Adds verify() cards for **S1Heisenberg1D/Susceptibility{XX,YY,ZZ}/OBC** — 3 uncorroborated hubs.

- Even N + Haldane phase ⇒ unique S_total = 0 GS ⇒ χ_αα = β·Var(S^α_total) = 0 exactly
- (Odd N has end-mode-doublet GS ⇒ χ ~ (2/3)·β/N, deferred.)

Card sweeps J ∈ {0.5, 1.0, 2.0}, N ∈ {4, 6}, all 3 axes. Refs #381.

### DIFF
--- a/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_susc_batch.jl
+++ b/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_susc_batch.jl
@@ -20,7 +20,7 @@ using QAtlas, Test
                     OBC(N);
                     route=:second_closed_form,
                     independent=0.0,
-                    agree_within=1e-9,
+                    agree_within=1e-8,
                     refs=["S1Heisenberg1D OBC even N: unique gapped Haldane S_total=0 GS ⇒ χ_αα = β·Var(S^α_total) = 0"],
                     fetch_kw=(; J=J, beta=LOW_T_BETA),
                 )
--- a/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_susc_batch.jl
+++ b/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_susc_batch.jl
@@ -1,0 +1,30 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_susc_batch.jl
+#
+# Susceptibility T → 0 limits for the OBC S=1 Heisenberg chain at even N:
+# the gapped Haldane GS at even N is a unique S_total = 0 singlet, so
+# χ_αα = β · Var(S^α_total) = 0 exactly.
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "S1Heisenberg1D — Susceptibility{XX,YY,ZZ}/OBC at even N, T→0 (#381 batch)" begin
+    LOW_T_BETA = 1e6
+    for J in (0.5, 1.0, 2.0)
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary

Adds 18 verify() cards (3 J × 2 N × 3 axes) for `S1Heisenberg1D/Susceptibility{XX,YY,ZZ}/OBC` at even N in the Haldane phase, claiming `χ_αα = 0` from the unique singlet GS. Physics correct; same tolerance-floor concern as PR #412.

### Findings

1. **Physics is correct.** The S=1 Heisenberg AFM chain on OBC at even N has a unique S_total = 0 GS deep in the Haldane phase (no spin-½ end modes can pair into a triplet at even N — the four-fold S=0,1 end-mode degeneracy only manifests with a non-zero coupling between ends, and at even N + OBC the lowest is the global singlet). Standard Affleck–Kennedy–Lieb–Tasaki / White–Huse numerical result. Static `χ_αα = β·Var(S^α_total)/N` collapses to 0 on the singlet. Independent value `0.0` is exact.

   *Caveat worth noting:* the "unique singlet GS at even N" statement is rigorous deep in the Haldane phase, but the four end-state quasi-degeneracy `(S=0,1)⊗2` becomes exponentially small with N. At N = 4 the splitting between S=0 and the lowest S=1 end-mode state is still O(1)·J due to the short chain — that's fine for this card because we only need the lowest level to be S=0, which it is. Just worth keeping in mind for future cards.

2. **Sweep coverage:** `J ∈ {0.5, 1.0, 2.0}` × `N ∈ {4, 6}` × 3 axes = 18 cards. N=6 for spin-1 is dim 3^6 = 729, healthy ED size. No N=8 (3^8 = 6561) — reasonable trade-off.

3. **Tolerance floor (same as PR #412).** With `β = 1e6`, the `β·Var(M)` formula has a cancellation-noise floor near `1e-9`. The chosen `agree_within = 1e-9` is exactly at that floor. Suggest `1e-8`:

   ```julia
   agree_within=1e-9,   # at β·eps floor; CI may flake on alt BLAS
   ```

4. **Route and refs.** `route=:second_closed_form` is correct for an analytical assertion on top of dense-ED. Refs string captures the mechanism cleanly.

5. **Test placement** at `test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_susc_batch.jl` matches the sibling `test_S1Heisenberg1D_*.jl` pattern (note minor casing inconsistency vs siblings using `S1Heisenberg1D` in filename — not blocking, but consider matching for consistency).

### Verdict

Approve modulo the tolerance bump to `1e-8`. Physics, route, and structure are sound. Optionally consider adding a comment that the singlet-uniqueness at even N relies on being deep in the Haldane phase, in case a future S=1 anisotropic extension reuses this pattern.

sotashimozono (commented):

Physics and code are correct.

**Physics:** S=1 OBC Heisenberg chain at even N is in the Haldane phase with a unique S_total=0 singlet ground state. The static susceptibility χ_αα = β·Var(S^α_total) vanishes identically because S^α_total|GS⟩ = 0. SU(2) symmetry makes the argument simultaneously valid for XX, YY, ZZ — no separate derivation needed per axis.

**Parameter choices:** J ∈ {0.5, 1.0, 2.0}, N ∈ {4, 6}, beta=1e6 are discriminating. The Boltzmann weight of the first excited state is exp(-Δ·1e6) ≈ 0 to machine precision for any non-zero Haldane gap Δ~0.4J, so `agree_within=1e-8` is correctly tight.

LGTM.

PR #416, opened 2026-05-20
Cards: XXZ1D SusceptibilityZZ/OBC T→0 (even + odd N) (#381)

sotashimozono:
Adds verify() card for **XXZ1D/SusceptibilityZZ/OBC** at T → 0 in both N parities.

- **Even N**: unique m_z = 0 singlet GS ⇒ Var(S^z_total) = 0 ⇒ χ_zz = 0
- **Odd N**: m_z = ±1/2 doublet GS ⇒ Var(σ^z_total) = 1 ⇒ χ_zz = β / N (σ-matrix convention)

Both hold for all Δ (U(1) S^z conservation). Card sweeps Δ ∈ {0.5, 1.0, 2.0}, even N ∈ {4,6,8} + odd N ∈ {3,5,7}. Refs #381.

### DIFF
--- a/test/models/quantum/XXZ/test_xxz1d_obc_susc_batch.jl
+++ b/test/models/quantum/XXZ/test_xxz1d_obc_susc_batch.jl
@@ -35,7 +35,7 @@ using QAtlas, Test
                 OBC(N);
                 route=:second_closed_form,
                 independent=BETA / N,
-                agree_within=1e-9,
+                agree_within=1e-6,  # odd-N branch: independent=β/N=1e5..3e5; relative ~3e-15 floor is borderline
                 refs=["XXZ1D OBC odd N: m_z=±1/2 doublet GS ⇒ χ_zz = β·Var(σ^z_total)/N = β/N"],
                 fetch_kw=(; J=J, Δ=Δ, beta=BETA),
             )
--- a/test/models/quantum/XXZ/test_xxz1d_obc_susc_batch.jl
+++ b/test/models/quantum/XXZ/test_xxz1d_obc_susc_batch.jl
@@ -1,0 +1,44 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/XXZ/test_xxz1d_obc_susc_batch.jl
+#
+# T → 0 susceptibility χ_zz of the OBC XXZ chain:
+#   * Even N: unique m_z = 0 singlet GS ⇒ Var(S^z_total) = 0 ⇒ χ_zz = 0
+#   * Odd N : m_z = ±1/2 doublet GS ⇒ Var(σ^z_total) = 1 ⇒
+#             χ_zz = β · Var / N = β / N   (σ-matrix normalisation)
+# Both branches hold for all Δ (by U(1) S^z conservation).
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "XXZ1D — SusceptibilityZZ/OBC T→0 (#381 batch)" begin
+    BETA = 1e6
+    for (J, Δ) in ((1.0, 0.5), (1.0, 1.0), (1.0, 2.0))
+        # Even N: χ_zz = 0
+        for N in (4, 6, 8)
+            verify(
+                XXZ1D(),
+                SusceptibilityZZ(),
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds a single `verify()` batch file `test/models/quantum/XXZ/test_xxz1d_obc_susc_batch.jl` covering `XXZ1D/SusceptibilityZZ/OBC` at T → 0 in both parities of N. The closed-form values come from the U(1) S^z-conservation structure of the GS sector. Pure verify(), no src changes.

### Findings

1. **σ-matrix convention is correct (verified against src).** `XXZ_thermal.jl` (header) explicitly states the susceptibility solver uses Pauli normalisation: `χ_αα = β Var(Σᵢ σᵅᵢ) / N`. With the m_z=±½ doublet (S convention) = ±1 (σ convention), Var(σ^z_total) = 1, so the claim `χ_zz = β/N` for odd N is exactly right. The `# σ-matrix normalisation` comment in the header matches the solver. Good.

2. **Even-N branch.** Unique m_z=0 singlet ground state ⇒ σ^z_total |GS⟩ = 0 ⇒ Var = 0 at T→0. The thermal kernel uses `eigen(Hermitian(H))` and weights are renormalised after an emin shift, so the excited-state leakage at β=1e6 is bounded by ~e^{-β·gap_per_chain}. For OBC XXZ at finite N, gap_per_chain ~ Δ/N ≳ 0.5/8 = 0.0625, so the leakage is far below `agree_within=1e-9`. Solid.

3. **Tolerance is borderline tight for the odd-N branch.** `independent = β/N = 1e6/N ≈ 1.3e5..3.3e5` with `atol = 1e-9` is effectively a relative tolerance of ~3e-15 — right at the floating-point boundary after a β·Var arithmetic. The thermal kernel multiplies by β, so an ε-level fluctuation in Var gives ~1e-10 absolute error. Likely passes but flaky if ED eigenvalues for the degenerate doublet drift by more than `1/β = 1e-6`. Consider loosening odd-N branch to `agree_within=1e-6` (still tight enough to catch a 1× vs 1/4× σ-vs-S convention mismatch, which would be a difference of ~10⁵).

4. **Parameter sweep.** Δ ∈ {0.5, 1.0, 2.0} covers gapless XXZ (|Δ|<1), Heisenberg (Δ=1), and gapped Néel (Δ>1). Missing the ferro regime Δ ≤ −1, where the GS is the polarised manifold and the m_z structure changes — but that's a separate physics regime, not required here. N ∈ {3,4,5,6,7,8} is good coverage.

5. **Route + refs.** `route = :second_closed_form` is the right tag (the parity-of-N argument is independent of the dense-ED computation). Refs strings are short but contain the actual derivation logic, which matches the established Cards #381 convention. Good.

6. **One small nit.** The two refs share the same physical fact (σ-vs-S convention dictates the prefactor). Could consolidate into one source. Cosmetic only.

### Verdict
**Approve with one suggestion.** Physics is correct; σ-convention matches the solver. The odd-N tolerance `1e-9` is borderline tight given the `β=1e6` magnification — recommend bumping to `1e-6` for that branch to avoid flake. Otherwise ready to merge.

sotashimozono (commented):

Physics and code are correct, with one nuance worth recording.

**Even-N branch (χ_zz = 0):** The unique m_z=0 singlet ground state at even N satisfies S^z_total|GS⟩ = 0, so Var(S^z_total) = 0 exactly and χ_zz = β·Var/N = 0. Holds for all Δ by U(1) symmetry.

**Odd-N branch (χ_zz = β/N):** The m_z=±1/2 doublet is equally populated at low T. Using σ-matrix normalization (σ^z_total eigenvalue ±1 for the two ground states), Var(σ^z_total) = 1, giving χ_zz = β·1/N = β/N. `independent=BETA / N` is consistent with this derivation.

**Tolerance for odd-N:** `agree_within=1e-6` in absolute terms is well-chosen — the value is O(β/N) = O(10^4 – 3×10^5), so the relative precision is ~10^{-11}, safely above machine epsilon. The comment explaining the denominator-scale reasoning is valuable for future maintainers.

**Parameter coverage:** Δ ∈ {0.5, 1.0, 2.0} with J=1.0 spans XY, SU(2), and Ising phases. N ∈ {4,6,8} (even) and {3,5,7} (odd) are sufficiently discriminating.

LGTM.

PR #417, opened 2026-05-20
Cards: Heisenberg1D VonNeumann+Renyi entropy OBC ℓ=1 = log 2 (#381)

sotashimozono:
Adds verify() cards for **Heisenberg1D/{VonNeumannEntropy, RenyiEntropy}/OBC** — 2 uncorroborated hubs.

SU(2) symmetry of the Heisenberg Hamiltonian forces the single-site reduced density matrix ρ_1 = I/2 (SU(2) averaging over orientations), so the single-site entanglement entropy is log 2 for all J, N, β:

- S_vN(ℓ=1) = log 2
- S_α(ℓ=1) = log 2 for all Rényi index α

Card sweeps J ∈ {0.5, 1.0, 2.0}, N ∈ {3,4,5,6}, β ∈ {0.5, 10, 10^6}, α ∈ {2, 3}. Refs #381.

PR #418, opened 2026-05-20
Cards: XXZ1D VonNeumann+Renyi entropy OBC ℓ=1 = log 2 (#381)

sotashimozono:
Adds verify() cards for **XXZ1D/{VonNeumannEntropy, RenyiEntropy}/OBC** — 2 uncorroborated hubs.

U(1) S^z conservation + xy-plane Z2 reflection ⇒ ρ_1 = I/2 (maximally mixed single-site reduced state) ⇒ S(ℓ=1) = log 2 for all J, Δ, N, β. Refs #381.

### DIFF
--- a/test/models/quantum/XXZ/test_xxz1d_obc_entropy_l1_batch.jl
+++ b/test/models/quantum/XXZ/test_xxz1d_obc_entropy_l1_batch.jl
@@ -2,7 +2,7 @@
 # test/models/quantum/XXZ/test_xxz1d_obc_entropy_l1_batch.jl
 #
 # Single-site entanglement entropy of the OBC spin-1/2 XXZ chain at
-# half-filling: U(1) z-rotation symmetry + xy-Z2 reflection forces the
+# half-filling: U(1) z-rotation × Z₂^x reflection (Z₂ from e^{iπΣS^x}) forces the
 # single-site reduced density matrix to be ρ₁ = I/2 (no preferred axis,
 # ⟨S^z⟩ = 0 in m_z=0 sector), so S_vN(ℓ=1) = S_α(ℓ=1) = log 2 for ALL
 # J, Δ, N, β.
@@ -22,7 +22,7 @@ using QAtlas, Test
                     route=:second_closed_form,
                     independent=log(2),
                     agree_within=1e-10,
-                    refs=["XXZ1D U(1)+xy-Z2 symmetry: ρ₁ = I/2 ⇒ S_vN(ℓ=1) = log 2 for all J, Δ, N, β"],
+                    refs=["XXZ1D U(1) × Z₂^x symmetry (Z₂ from e^{iπΣS^x}): ρ₁ = I/2 ⇒ S_vN(ℓ=1) = log 2 for all J, Δ, N, β"],
                     fetch_kw=(; J=J, Δ=Δ, ℓ=1, beta=β),
                 )
                 for α in (2, 3)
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds `verify()` cards for `XXZ1D/{VonNeumannEntropy, RenyiEntropy}/OBC` at ℓ=1, independent value `log 2`, derived from `U(1) ⊗ ℤ₂` (xy-flip) of the XXZ Hamiltonian. Pure verify(), no src changes.

### Findings

1. **Symmetry argument is correct.** XXZ has full SU(2) only at Δ=1, but U(1) S^z-conservation + global ℤ₂ (e.g. e^{iπ Σ S^x_i}, sending σ^z → −σ^z) is preserved for **all** Δ ∈ ℝ. U(1) forces the off-diagonal entries of ρ_1 (in σ^z basis) to vanish; ℤ₂ forces p_↑ = p_↓ ⇒ ρ_1 = I/2 ⇒ S = log 2. Holds for any β and any sign of Δ. ✓

2. **Δ = −0.5 case.** Tested explicitly with `(J, Δ) = (2.0, -0.5)`. The ferro/Δ<0 side keeps U(1) and ℤ₂ symmetry, so the prediction still holds. Good — flushes out anyone tempted to assume the result is AFM-only. Note: `|Δ| ≥ 1` (gapped Néel / ferro) is **not** in the sweep — fine because the symmetry argument is Δ-independent, but adding Δ = 2 (gapped) would have made the universality more obvious in the card.

3. **Hub dispatch verified.** `XXZ_thermal.jl::fetch(XXZ1D, VonNeumannEntropy, OBC; ℓ, beta=Inf, kwargs...)` and the `RenyiEntropy` sibling both exist with the expected signature; the partial-trace helper `_xxz1d_partial_trace_B(ρ, ℓ, N)` handles ℓ=1 cleanly. Numerical path is `eigvals(Hermitian(ρA))`, capped at `_MAX_ED_SITES = 12`. The test uses N ≤ 6 — well under the cap.

4. **β sweep.** β ∈ {0.5, 10, 1e6} covers high-T → effective GS. Same as PR #417 — the U(1)+ℤ₂ argument doesn't care about T, so the temperature sweep is exactly the right falsification probe.

5. **One subtlety worth a one-liner in the refs.** The ℤ₂ here is the "global spin flip" e^{iπ Σ S^x}, not the bond-level Z₂. The current ref string says "xy-Z2 symmetry" which is mildly ambiguous (could be misread as the Z₂ that maps ↑↔↓ on a single site, which is not a symmetry by itself). Suggest tightening to: `"global ℤ₂ spin-flip (π-rotation about x): σᶻ → −σᶻ"`. Cosmetic.

6. **Tolerance.** `1e-10` is fine; the eigvals of a 2×2 PSD matrix that should be diag(½,½) are reproduced to ~1e-15 by `eigvals(Hermitian(...))`.

7. **route = `:second_closed_form`.** Correct — the symmetry argument is structurally distinct from the dense-ED partial trace.

### Verdict
**Approve.** Symmetry reasoning is correct and Δ-independent (test correctly probes Δ<0 too). The ref string could clarify which ℤ₂ is meant — a one-character edit but not blocking. Merge as-is.

sotashimozono (commented):

Physics and code are correct.

**Physics:** For the S=1/2 OBC XXZ chain, U(1) S^z conservation forces ρ₁ to be diagonal in the σ^z eigenbasis. At half-filling (working in the m_z=0 total-spin sector), the Z₂ symmetry e^{iπΣS^x} maps m_z → -m_z within the sector, forcing ⟨σ^z_i⟩ = 0, so the diagonal elements of ρ₁ are (1/2, 1/2), giving ρ₁ = I₂/2. Entropy S_vN(ℓ=1) = S_α(ℓ=1) = log 2 for all Rényi orders.

One subtlety: at Δ=2.0 (Ising phase) the infinite-volume ground state spontaneously breaks Z₂, but in finite OBC the symmetry is never broken and ρ₁ = I₂/2 still holds exactly — the card correctly covers this case.

**Coverage:** (J, Δ) ∈ {(1.0,0.5), (1.0,1.0), (1.0,2.0), (2.0,-0.5)} spans XY phase, SU(2) point, Ising phase, and easy-plane; N ∈ {3,4,5,6}; β ∈ {0.5, 10, 10^6}; α ∈ {2,3}. Discriminating.

**Tolerances:** `agree_within=1e-10` is appropriate.

LGTM.

PR #421, opened 2026-05-20
Cards: TFIM LoschmidtEcho rate function at t=0 = 0 (#381)

sotashimozono:
Adds verify() cards for **TFIM/{VonNeumannEntropy, RenyiEntropy}/OBC** — 2 uncorroborated hubs.

- **J = 0** at T → 0: GS = |+⟩^N pure product state ⇒ ρ_1 pure ⇒ S(ℓ=1) = 0
- **h = 0** at any β: Z2 symmetry σ_z → −σ_z and no x-channel ⇒ ρ_1 = I/2 ⇒ S(ℓ=1) = log 2

Both branches verified for VonNeumann and Rényi α ∈ {2,3}. Refs #381.

### DIFF
--- a/test/models/quantum/TFIM/test_TFIM_loschmidt_t0_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_loschmidt_t0_batch.jl
@@ -1,0 +1,42 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/TFIM/test_TFIM_loschmidt_t0_batch.jl
+#
+# Loschmidt-echo rate function r(t) = -(1/N) log |⟨ψ₀|e^{-iHt}|ψ₀⟩|² at
+# t = 0: the time-evolution operator is identity, so ⟨ψ₀|ψ₀⟩ = 1 and
+# r(0) = 0 identically, for ANY initial-final TFIM pair.
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "TFIM — LoschmidtEcho rate function at t=0 = 0 (#381 batch)" begin
+    # /Infinite
+    for (J, h_i, h_f) in (
+        (1.0, 0.5, 1.0), (1.0, 0.5, 2.0), (1.0, 1.5, 1.0), (1.0, 1.5, 2.0),
+        (2.0, 0.5, 3.0), (0.5, 2.0, 1.0),
+    )
+        verify(
+            TFIM(; J=J, h=h_f),
+            LoschmidtEcho(),
+            Infinite();
+            route=:second_closed_form,
+            independent=0.0,
+            agree_within=1e-10,
+            refs=["Loschmidt rate function r(t) = -(1/N) log |⟨ψ₀|e^{-iHt}|ψ₀⟩|² ⇒ r(t=0) = 0 (identity evolution, normalized state)"],
+            fetch_kw=(; initial=TFIM(; J=J, h=h_i), t=0.0),
+        )
+        # /OBC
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds a `verify()`-only batch asserting that the Loschmidt-echo rate function r(t) = -(1/N) log |⟨ψ₀|e^{-iHt}|ψ₀⟩|² vanishes at t=0 for TFIM quenches across `Infinite()` and `OBC(N)` boundary conditions, for several (J, h_i, h_f) combinations.

### Findings

1. **The identity is trivially true by definition** — at t=0 the evolution operator is the identity, ⟨ψ₀|ψ₀⟩ = 1 for any normalized initial state, so log(1) = 0 regardless of model, BC, quench parameters, or N. The card therefore exercises the *plumbing* (parameter wiring, BC dispatch, `initial=` kwarg, hub registration) rather than physics. That is a legitimate use of `verify()` as a smoke test for the LoschmidtEcho hub on TFIM.

2. **Tolerance choice (1e-10) is right** for a value that should be exactly 0 modulo floating-point noise on `-log(|1+ε|²)/N`.

3. **Quench coverage is thoughtfully chosen** — same-phase (h=0.5→1.0 both PM-side of h_c=1), cross-critical (0.5↔2.0, 1.5↔1.0, 1.5↔2.0), and J-rescaling (J=2, J=0.5) cases. Since the result is t=0, none of this actually stresses the dynamics, but it does confirm the hub accepts these parameter shapes.

4. **`Infinite()` branch is interesting** — the rate function uses 1/N normalization, so passing `Infinite()` implicitly relies on the hub computing a per-site quantity that has a well-defined thermodynamic-limit value (which is 0 here by triviality). Worth confirming the underlying solver actually returns 0.0 rather than `NaN` from a 1/N → 1/∞ path; the tight 1e-10 tolerance will catch any such issue.

5. **Minor: stray `# /Infinite` comment** sits awkwardly on its own line at the top of the inner loop. The pattern `# /Infinite` then later `# /OBC` is fine but the first one being separated from its `verify()` by nothing is mildly confusing. Cosmetic only.

6. **Missing follow-up sibling card (informational, not blocking):** the natural companion is r(t) for the *non-quenched* case (h_i = h_f), where r(t) should remain 0 for all t (initial state is an eigenstate, |⟨ψ₀|e^{-iEt}|ψ₀⟩|² = 1). That would actually probe the time-evolution path and isn't redundant with this trivial t=0 card.

### Verdict
Approve. Trivial identity, but a useful smoke test for the LoschmidtEcho hub across BCs. Consider the t>0 trivial-eigenstate companion as a follow-up.

sotashimozono (commented):

Physics and code are correct.

**Physics:** The Loschmidt-echo rate function is r(t) = -(1/N) log |⟨ψ₀|e^{-iHt}|ψ₀⟩|². At t=0, e^{-iHt} = I, so ⟨ψ₀|ψ₀⟩ = 1 (normalized state), giving r(0) = -(1/N) log 1 = 0. This holds for any initial state |ψ₀⟩, any final Hamiltonian H, any boundary condition, and any system size — it is a purely kinematic identity.

**Parameter choices:** Six distinct (J, h_i, h_f) quench pairs including post-critical (h_f=2.0>J=1.0) and pre-critical (h_f=0.5<J=1.0) protocols, plus N ∈ {8,12} for OBC. Crossing the quantum critical point h_c=J in the quench parameters is a good discriminating choice — the t=0 triviality must hold even when the quench is across a DQPT.

**Tolerances:** `agree_within=1e-10` is appropriate for an exact identity.

The follow-up suggestion (non-trivial h_i ≠ h_f quench at finite t probing DQPT zeros) is a natural next card — worth tracking in #381.

LGTM.

PR #422, opened 2026-05-20
Cards: TFIM {XX,YY,ZZ}StructureFactor T→∞ paramagnet = 1 (#381)

sotashimozono:
Adds verify() cards for **TFIM/{VonNeumannEntropy, RenyiEntropy}/OBC** — 2 uncorroborated hubs.

- **J = 0** at T → 0: GS = |+⟩^N pure product state ⇒ ρ_1 pure ⇒ S(ℓ=1) = 0
- **h = 0** at any β: Z2 symmetry σ_z → −σ_z and no x-channel ⇒ ρ_1 = I/2 ⇒ S(ℓ=1) = log 2

Both branches verified for VonNeumann and Rényi α ∈ {2,3}. Refs #381.

### DIFF
--- a/test/models/quantum/TFIM/test_TFIM_structure_factor_paramagnet_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_structure_factor_paramagnet_batch.jl
@@ -5,6 +5,15 @@
 # ρ = I / 2^N, which gives ⟨σ^α_i σ^α_j⟩ = δ_{ij} (uncorrelated) and
 # therefore S_αα(q) = Σ_r e^{iqr} ⟨σ^α_0 σ^α_r⟩ → 1 for all q, all axes,
 # any BC. Pure verify(); branches off main. Refs #381.
+#
+# TODO(#438): This card uses the σ-matrix convention (⟨σ^α⟩² = 1 ⇒
+# independent = 1.0). Under the project-wide spin-S convention policy
+# being rolled out in PR #438 (docs/src/conventions.md), TFIM observable
+# returns must be reported in S = σ/2 units, which would change the
+# expected paramagnet value to ⟨S^α_i S^α_j⟩|_{i=j} = 1/4. Migrate this
+# file (independent = 0.25, tighten agree_within to 1e-9) once #438
+# merges; until then the σ convention is internally consistent with the
+# current TFIM structure-factor implementation.
 # ─────────────────────────────────────────────────────────────────────────────
 
 using QAtlas, Test
--- a/test/models/quantum/TFIM/test_TFIM_structure_factor_paramagnet_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_structure_factor_paramagnet_batch.jl
@@ -1,0 +1,45 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/TFIM/test_TFIM_structure_factor_paramagnet_batch.jl
+#
+# At T → ∞ the TFIM thermal state is the maximally mixed state
+# ρ = I / 2^N, which gives ⟨σ^α_i σ^α_j⟩ = δ_{ij} (uncorrelated) and
+# therefore S_αα(q) = Σ_r e^{iqr} ⟨σ^α_0 σ^α_r⟩ → 1 for all q, all axes,
+# any BC. Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds a `verify()`-only batch asserting that the TFIM {XX, YY, ZZ} structure factor reduces to a constant (target = 1) in the T→∞ paramagnetic limit, for q ∈ {0, π/2, π} across `Infinite()` and `OBC(N=8,12)`.

### Findings

1. **The target value 1 implies a specific normalization convention** — check the file matches. For Pauli operators σ^α at T→∞:
   ⟨σ^α_i σ^α_j⟩ = δ_{ij} · ⟨(σ^α)²⟩ = δ_{ij} · 1 (since σ² = I).
   With `S(q) = Σ_r e^{iqr} ⟨σ^α_0 σ^α_r⟩` (no 1/N prefactor) only the r=0 term survives, giving 1. With `S(q) = (1/N) Σ_{i,j} e^{iq(r_i−r_j)} ⟨σ^α_i σ^α_j⟩` you also get (1/N)·N·1 = 1. If instead the convention uses spin operators S^α = σ^α/2, the diagonal piece becomes ⟨(S^α)²⟩ = 1/4 and the target should be 1/4, not 1. **Worth a one-line comment in the file pinning down which operator convention (σ vs S) and which prefactor (Σ vs (1/N)Σ) the hub implements**, so the next person reading the card knows why 1 is the right answer.

2. **`HIGH_T_BETA = 1e-3` plus `agree_within = 1e-2` is the right tradeoff,** but tightening the temperature would tighten the bound. Leading correction at small β is O(β²·J·h) (first order O(β) vanishes by σ_z ↔ −σ_z symmetry in ZZ; for XX/YY it requires a careful expansion). 1e-2 is safe; if you want a sharper card, β=1e-4 with agree_within=1e-3 should still pass.

3. **`Infinite()` branch** — at β=1e-3 the high-T expansion converges fast and the per-site result is N-independent, so passing `Infinite()` is fine. Make sure the hub's `Infinite()` solver doesn't internally pick a default N that's so small that finite-size kicks in above 1e-2; the OBC(8,12) cases below act as a partial check (they agree means N≥8 is enough).

4. **Q-coverage** {0, π/2, π} captures all three symmetry-distinct momenta on a lattice, which is the right minimal set.

5. **Axis coverage** {XX, YY, ZZ} is complete. TFIM lacks rotational symmetry in spin space (h breaks SU(2) and even U(1)), so the three components are independently verified — good.

6. **Minor:** the docstring says `S_αα(q) = Σ_r e^{iqr} ⟨σ^α_0 σ^α_r⟩ → 1 for all q`. This is correct *for the σ-operator, no-prefactor* convention. If the actual hub differs (see point 1), update the docstring; otherwise good.

### Verdict
Approve, with the request to pin down the normalization convention in the header comment so the "= 1" target is unambiguous to future readers. Tolerance and parameter coverage are well-chosen.

sotashimozono (commented):

Physics and code are correct. One convention note is properly handled.

**Physics:** At T→∞ the TFIM thermal state is ρ = I/2^N (maximally mixed). For any two sites i≠j: ⟨σ^α_i σ^α_j⟩ = Tr[ρ σ^α_i σ^α_j] = Tr[σ^α_i/2] · Tr[σ^α_j/2] = 0 (traceless). For i=j: ⟨(σ^α_i)²⟩ = 1 (Pauli matrices square to identity). Therefore the structure factor S_αα(q) = (1/N) Σ_{i,j} e^{iq(i-j)} ⟨σ^α_i σ^α_j⟩ = (1/N) · N · 1 = 1, independent of q, α, J, h, and BC.

**Convention handling:** The TODO comment about the upcoming σ→S spin convention migration (PR #438) is a good practice — the card explicitly documents the current σ-convention assumption (`independent=1.0`) and the future migration target (`independent=0.25` after dividing by 4). This prevents silent breakage when #438 lands.

**Tolerances:** `agree_within=1e-2` is appropriate — beta=1e-3 gives T=1000 which is a good approximation to T=∞ but not exact, so a loose tolerance is correct. The route `:limiting_case` (rather than `:second_closed_form`) is the right choice for a high-T approximation.

**Coverage:** (J, h) ∈ {(1.0,0.5), (1.0,1.0), (1.0,2.0), (0.5,2.0)}; XX, YY, ZZ structure factors; q ∈ {0, π/2, π}; Infinite and OBC N ∈ {8,12}. The q-independence is well-tested.

LGTM.

PR #424, opened 2026-05-20
Cards: S1Heisenberg1D {XX,YY,ZZ}Correlation/OBC at i=j = 2/3 SU(2) (#381)

sotashimozono:
Adds verify() cards for **S1Heisenberg1D/{XX,YY,ZZ}Correlation/OBC** at equal sites i=j — 3 uncorroborated hubs.

SU(2) symmetry of the S=1 Heisenberg Hamiltonian ⇒ ρ_i = I/3 ⇒ ⟨(S^α)²⟩ = (1/3) Tr(S_α²) = S(S+1)/3 = 2/3, axis- and β-independent.

Card sweeps J ∈ {0.5,1,2}, N ∈ {3,4,5,6}, β ∈ {0.5, 10, 10^6}, all 3 axes. Refs #381.

### DIFF
--- a/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_corr_eqsite_batch.jl
+++ b/test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_corr_eqsite_batch.jl
@@ -3,9 +3,12 @@
 #
 # Equal-site spin-spin correlator of the OBC spin-1 Heisenberg chain:
 #   ⟨S^α_i S^α_i⟩_β = ⟨(S^α)²⟩_β
-# By SU(2) symmetry the single-site reduced state is ρ_i = I/3, so
-# ⟨(S^α)²⟩ = (1/3) Tr(S_α²) = S(S+1)/3 = 2/3 for S=1, independent of
-# the axis α, J, N, β.
+# By SU(2) covariance of the thermal ensemble + spin-1 Schur's lemma:
+# the rank-2 tensor T^{αβ}_i ≡ ⟨S^α_i S^β_i⟩_β commutes with global SU(2),
+# so T^{αβ}_i ∝ δ^{αβ} with trace Σ_α T^{αα}_i = ⟨S_i·S_i⟩ = S(S+1) = 2,
+# hence ⟨S^α_i S^α_i⟩ = ⟨(S^α)²⟩ = (1/3)·S(S+1) = 2/3 for S=1, at every
+# site i, axis α, J, N, β. (Note: this does NOT require ρ_i = I/3 — that
+# stronger statement fails at finite-N OBC edge sites in the Haldane regime.)
 # Pure verify(); branches off main. Refs #381.
 # ─────────────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds a `verify()`-only batch asserting ⟨S^α_i S^α_i⟩ = 2/3 for the S=1 Heisenberg chain (OBC) at any (J, N, β, axis α), justified by SU(2) symmetry. Covers J ∈ {0.5,1,2}, N ∈ {3,4,5,6}, β ∈ {0.5, 10, 10^6}, all three axes.

### Findings

1. **SU(2) argument is correct, but the stated derivation in the header is sloppy.** The header says "ρ_i = I/3, so ⟨(S^α)²⟩ = (1/3) Tr(S_α²) = S(S+1)/3 = 2/3". The claim **ρ_i = I/3 is too strong for finite N OBC GS** — the S=1 Haldane chain has nearly-degenerate edge modes at finite N (effective edge spin-1/2 pair giving a quasi-fourfold GS multiplet split exponentially in N). For finite N and modest β you are *not* in a pure singlet, so the single-site marginal is not literally I/3 at every site in a single GS.
   The correct, cleaner argument is: the thermal density matrix ρ = e^{−βH}/Z commutes with every S^α_total because [H, S^α_total] = 0 for SU(2)-invariant H. Therefore the two-index tensor T^{αβ} ≡ ⟨S^α_i S^β_i⟩_thermal is SU(2)-covariant on a scalar, hence proportional to δ^{αβ}, with trace Σ_α T^{αα} = ⟨S_i·S_i⟩ = S(S+1) = 2 ⇒ each diagonal = 2/3. This holds **regardless of whether ρ_i is I/3** — it only requires SU(2)-covariance of the thermal ensemble of a two-spin operator. Please update the doc-comment; the current "ρ_i = I/3" line will mislead the next reader (especially since it's false at edge sites of a finite OBC chain in some sectors).

2. **Boundary / edge sites:** the card uses `i=1, j=1` — an edge site. The fix above guarantees this is still 2/3 by the SU(2) covariance argument applied to that fixed site (the same calculation works at any single site). Good choice for stress-testing (edge sites are where naive intuition most often fails); just make sure the doc reflects that we're not appealing to translation invariance.

3. **β-coverage** spans high-T (0.5), low-T (10), and effectively GS (1e6). All three should give 2/3 exactly by the argument in finding 1. Tight tolerance `agree_within=1e-8` is appropriate — the only sources of error are exponential-diagonalization rounding and partition-function summation, both well within 1e-8.

4. **N=3 (odd, AFM S=1)** is a particularly nice test — odd-length SU(2) chains have total-spin sectors that can confuse naive product-state pictures, and the SU(2)-covariance argument cuts through that cleanly. Good inclusion.

5. **Hub coverage:** all three of {XX, YY, ZZ}Correlation with `mode=:static`. By the same SU(2) argument the three are forced equal to each other, so this card also implicitly checks that the solver respects SU(2) symmetry to 1e-8 (would catch a subtle bug in the H construction or Sz-quantization-axis bias).

6. **Minor:** no `i=2`/`i=middle` companion. Since the SU(2) argument doesn't depend on site, this is fine; but a one-liner adding `i=cld(N,2), j=cld(N,2)` would catch any solver bug that happens to be edge-site-only. Optional follow-up.

### Verdict
Approve, but please tighten the header argument to use SU(2)-covariance of the thermal ensemble rather than the "ρ_i = I/3" line, which is misleading for finite-N OBC. Tolerance, axis coverage, and (J, N, β) sweep are all well chosen.

sotashimozono (commented):

**Physics correctness**: The identity ⟨S^α_i S^α_i⟩ = S(S+1)/3 = 2/3 for S=1 is exact. The argument in the file comment is correct: the rank-2 tensor T^{αβ}_i = ⟨S^α_i S^β_i⟩ must be proportional to δ^{αβ} by Schur's lemma applied to the global SU(2)-covariance of the thermal state, independently of N, J, β, boundary condition, or site index. Trace S(S+1) = 2 → each diagonal component 2/3. The clarification that this does NOT require ρ_i = I/(2S+1) (which fails at finite-N OBC edge sites in the Haldane regime) is an important precision that goes beyond the naive argument.

**Code**: Correct use of verify() with route=:second_closed_form. Tolerance agree_within=1e-8 is appropriate for this analytically exact result. Batch covers J ∈ {0.5,1,2}, N ∈ {3–6}, β ∈ {0.5,10,1e6}, all three axes (XX,YY,ZZ), i=j=1. Good that i=1 is an OBC edge site — the most physically non-trivial case.

PR #350, opened 2026-05-13
feat(quantum): Hubbard1D Phase 2 — LuttingerParameter K=1 at U=0 free-fermion limit

sotashimozono:
## Summary

Phase-2 follow-up to **Hubbard1D** (1D Hubbard model, Lieb–Wu 1968): exposes the **Luttinger-liquid parameter** in the **free-fermion limit** U = 0.

- `fetch(::Hubbard1D, ::LuttingerParameter, ::Infinite)` returns `K = 1.0` at U = 0 (non-interacting spinful fermions, SU(2)-symmetric charge & spin sectors; Voit 1995 TLL review).
- Any U ≠ 0 raises `DomainError`: finite-U Lieb–Wu Bethe-ansatz K_ρ(U), K_σ(U) integrals are non-closed-form and deferred to a later Phase 2/3.
- Matches `TightBinding1D` (PR #316), which also returns K = 1 for free fermions — keeping `LuttingerParameter` universally consistent across free-fermion limits.

## Files

- `src/models/quantum/Hubbard1D/Hubbard1D.jl` — new fetch method (docstring + DomainError gating).
- `src/models/quantum/Hubbard1D/Hubbard1D_registry.jl` — new `@register` row (`method=:analytic`, `reliability=:high`).
- `test/models/quantum/misc/test_hubbard1d.jl` — 2 new testsets (6 tests).

`src/core/quantities.jl` and `src/QAtlas.jl` are **not** touched (`LuttingerParameter` already exported).

## Test plan

- [x] K == 1.0 at default `Hubbard1D(; U=0.0)`.
- [x] t-independence: K = 1 for t ∈ {0.5, 1.0, 3.0} at U = 0.
- [x] DomainError for U = 4 and U = -1.
- [x] Local run on Panza: 23 (Phase 1) + 4 + 2 = 29 tests pass.
- [ ] CI green on PR.

## References

- E. H. Lieb, F. Y. Wu, *Phys. Rev. Lett.* **20**, 1445 (1968).
- J. Voit, *Rep. Prog. Phys.* **58**, 977 (1995) — TLL review for Hubbard.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #425, opened 2026-05-20
Cards: Hubbard1D LuttingerParameter/Infinite at U=0 = 1 (#381)

sotashimozono:
Adds verify() card for **Hubbard1D/LuttingerParameter/Infinite** at U=0 (free-fermion limit).

At U=0 the Hubbard model at half-filling reduces to free spinful fermions ⇒ Luttinger parameter K = 1 (free-boson value), exact.

Off U=0 the parameter requires the Voit 1995 Lieb-Wu integral form (non-closed-form, src Phase 2 — currently throws DomainError). Refs #381.

### DIFF
--- a/test/models/quantum/misc/test_hubbard1d_luttinger_U0_batch.jl
+++ b/test/models/quantum/misc/test_hubbard1d_luttinger_U0_batch.jl
@@ -18,7 +18,7 @@ using QAtlas, Test
             route=:second_closed_form,
             independent=1.0,
             agree_within=1e-14,
-            refs=["At U=0 the Hubbard model is a free spinful fermion at half-filling ⇒ Luttinger parameter K = 1 (free-boson value)"],
+            refs=["At U=0 the Hubbard model is a free spinful fermion at half-filling (μ = U/2 = 0) ⇒ both Luttinger parameters K_ρ = K_σ = 1 (free-boson value, spin-charge separated free chains)"],
         )
     end
 end
--- a/test/models/quantum/misc/test_hubbard1d_luttinger_U0_batch.jl
+++ b/test/models/quantum/misc/test_hubbard1d_luttinger_U0_batch.jl
@@ -1,0 +1,24 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/misc/test_hubbard1d_luttinger_U0_batch.jl
+#
+# Hubbard1D Luttinger parameter at U = 0 (free spinful fermion at half
+# filling): K = 1 exactly.  Off U = 0 the model requires the Voit 1995
+# Lieb–Wu integrals (Phase 2, not closed-form).
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "Hubbard1D — LuttingerParameter/Infinite at U=0 = 1 (#381 batch)" begin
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds a `verify()` card asserting that Hubbard1D `LuttingerParameter/Infinite` equals 1 at U=0, swept over `t ∈ {0.5, 1.0, 2.0}` at half filling (`μ=0.0`). Pure exact-limit corroboration via `route=:second_closed_form` with `independent=1.0`, `agree_within=1e-14`. Small, scoped, branches off main per the #381 batch pattern.

### Findings

1. **Physics is correct but the card silently picks one of two K's (charge/spin) — worth a one-line ref note.** At the SU(2)-symmetric free-fermion point both K_ρ and K_σ equal 1, so the assertion `K = 1` is right whichever sector the hub returns. *But*: the U→0 limit is one-sided in K_ρ (K_ρ → 1⁻ from the repulsive side, K_ρ → 1⁺ from the attractive side; the spin Luttinger parameter is pinned to K_σ = 1 by SU(2) wherever it exists). At **strict** U=0 the answer is the free-fermion 1 exactly, so the card itself is fine — but the `refs` line would be more honest as: *"At U=0 (free spinful fermion at half-filling, SU(2)-symmetric) K_ρ = K_σ = 1 exactly; one-sided U → 0± limits are covered by the Voit (1995) integral form, not this card."* That documents which K the hub is reporting and warns the next reader off using this card to argue anything about the U → 0⁺ direction.

2. **Half-filling is hardcoded but not flagged.** `μ=0.0` (half filling) is essential to the scope; the ref should mention "half filling, SU(2)-symmetric" so a future reader does not generalise this card to arbitrary doping.

3. **Tolerance `1e-14` is appropriate** for a strictly closed-form / second-closed-form path that should not touch any numerical solver. If the path *does* route through any quadrature, this will catch it — good.

4. **Coverage: 3 × t values is sufficient** because the answer is t-independent at U=0 (the bandwidth drops out of K). Even one `t` value would suffice physics-wise; sweeping 3 is cheap insurance against a buggy t-dependence.

5. **No issue with `route=:second_closed_form`** — the U=0 branch is a closed-form limit of the Voit/Lieb-Wu expression and matches the prior PR #350 phase-2 work.

### Verdict
**Approve.** Card is correct, tight, and aligned with the #381 batch convention. The one suggested change is a one-line refinement of the `refs` string to disambiguate K_ρ vs K_σ and pin "half filling, SU(2)-symmetric" — optional, non-blocking.

sotashimozono (commented):

**Physics correctness**: K_ρ = 1 at U=0 is exact. At U=0 the 1D Hubbard model at half-filling separates into two independent spinless tight-binding chains (spin-↑ and spin-↓), each realizing a free Luttinger liquid with K = 1. Interactions shift K_ρ: repulsive U > 0 gives K_ρ < 1, attractive U < 0 gives K_ρ > 1. Setting μ = U/2 = 0 at U=0 correctly enforces half-filling. The refs string correctly mentions both K_ρ and K_σ = 1 (the spin sector is also a free boson at U=0).

**Code**: agree_within=1e-14 is tight and appropriate here — this is a pure closed-form result with no numerical integration involved. Batch over t ∈ {0.5,1,2} verifies the t-independence of K (dimensionless). No fetch_kw needed since U=0 is set in the model constructor.

PR #426, opened 2026-05-20
Cards: TFIM FidelitySusceptibility/OBC at J=0 = 0 (#381)

sotashimozono:
Adds verify() card for **TFIM/FidelitySusceptibility/OBC** at J=0.

At J=0 the TFIM Hamiltonian H = -h Σ σ_x has GS = |+⟩^N independent of h (only the energy rescales with h, the GS direction is fixed) ⇒ χ_F = 0 exactly. Refs #381.

### DIFF
--- a/test/models/quantum/TFIM/test_TFIM_fidelity_susc_J0_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_fidelity_susc_J0_batch.jl
@@ -1,10 +1,12 @@
 # ─────────────────────────────────────────────────────────────────────────────
 # test/models/quantum/TFIM/test_TFIM_fidelity_susc_J0_batch.jl
 #
-# At J = 0 the TFIM Hamiltonian H = -h Σ σ_x has GS = |+⟩^N independent of h
-# (only the energy depends on h, not the eigenstate).  Changing h is a pure
-# rescaling that does not rotate the GS, so the fidelity susceptibility
-# χ_F = |⟨∂_h ψ_0|∂_h ψ_0⟩_⊥|² vanishes identically.
+# At J = 0 the TFIM Hamiltonian H = -h Σ σ_x has GS = |+⟩^N (trivial
+# paramagnet) independent of h — only the energy depends on h, not the
+# eigenstate.  The fidelity susceptibility χ_F here is computed with
+# respect to the transverse field h (∂_h-derivative); since ⟨ψ(h)|ψ(h+dh)⟩ = 1
+# identically, χ_F(∂_h) = 0 exactly.  Note: this is NOT the same as ∂_J χ_F,
+# which is non-zero at J = 0.
 # Pure verify(); branches off main. Refs #381.
 # ─────────────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds a `verify()` card asserting that TFIM `FidelitySusceptibility/OBC` equals 0 at J=0, swept over `h ∈ {0.5, 1.0, 2.0}` and `N ∈ {4, 8, 12}`. Uses `route=:second_closed_form`, `independent=0.0`, `agree_within=1e-12`.

### Findings

1. **Physics argument is correct, and notably the right argument is the one in the file (not the naive one).** χ_F at J=0 is the second derivative of fidelity F(g, g+δg) where g is the parameter being varied. The PR body and file header correctly identify that *the GS = |+⟩^N is h-independent at J=0* — only the energy rescales with h, the GS direction is fixed — so |∂_h ψ_0⟩ = 0 and χ_F = 0 exactly. This is the singularity-free trivial limit; the QPT is at h = J (which is well away from this card's slice).

2. **One subtle clarification worth adding to `refs`:** at strict J=0 the GS is `|+⟩^N` *only as long as h > 0* (sign convention: H = -h Σ σ^x makes |+⟩^N the unique GS for h > 0, |−⟩^N for h < 0, and at h=0 the spectrum is 2^N-fold degenerate). The sweep `h ∈ {0.5, 1.0, 2.0}` correctly avoids h=0 — but stating "for any h > 0" in the ref would make the limit explicit and prevent a future contributor from adding h=0 to the sweep and getting a non-zero χ_F due to degenerate-GS ambiguity.

3. **Which susceptibility — w.r.t. h or w.r.t. J?** The card name `FidelitySusceptibility` is ambiguous about the parameter being differentiated. The physics argument in the file uses ∂_h, which makes χ_F = 0 because |+⟩^N is h-independent. **If the hub instead computes ∂_J**, the same conclusion still holds at J=0 because the perturbation +δJ Σ σ^z σ^z lifts degeneracies in a way that does shift |ψ_0⟩, so ∂_J|ψ_0⟩ ≠ 0 and χ_F(J=0) would be nonzero in that case. Please confirm the hub's differentiation variable matches the proof. If it's ∂_J, the card's `independent=0.0` is wrong; if it's ∂_h (as the file header assumes) the card is correct. A `refs` line stating *which variable* the susceptibility is taken with respect to is essential before merge.

4. **Coverage** (3 × 3 = 9 cards) is good — varies both `h` and `N`. If the hub size-extrapolates `Infinite`, only one N would be needed; for OBC ED, sweeping N is the right call and catches any finite-size leakage.

5. **Tolerance `1e-12`** is sensibly looser than 1e-14 because OBC ED may involve a finite-difference derivative that introduces machine-epsilon × stencil noise. Reasonable.

### Verdict
**Approve with one clarifying question** — please confirm point 3 (∂_h vs ∂_J) and add the differentiation variable to the `refs` string. If the hub differentiates w.r.t. h the card ships as-is; if it differentiates w.r.t. J the expected value is *not* 0 at J=0 and the card needs to be rescoped. One-line fix either way.

sotashimozono (commented):

**Physics correctness**: At J=0 the TFIM reduces to H = -h Σ_i σ^x_i, a sum of independent single-site terms. The ground state is |+⟩^⊗N = (|↑⟩+|↓⟩)^⊗N / 2^{N/2}, which is exactly the +1 eigenstate of each σ^x_i and is independent of h (only the GS energy -hN changes). The fidelity susceptibility with respect to h is χ_F = lim_{δh→0} -2 ln|⟨ψ(h)|ψ(h+δh)⟩|/δh² = 0, because the eigenstate does not change at all. The file comment correctly distinguishes ∂_h χ_F (zero, tested here) from ∂_J χ_F (non-zero), which is the right disambiguation to make.

**Code**: agree_within=1e-12 is appropriate for an identically-zero result via closed form. Batch over h ∈ {0.5,1,2}, N ∈ {4,8,12} provides good coverage. J=0.0 is passed in the model constructor.

PR #427, opened 2026-05-20
Cards: TFIM XXCorrelation/Infinite trivial limits (#381)

sotashimozono:
Adds verify() card for **TFIM/XXCorrelation/Infinite** at two trivial limits.

- **i = j**: ⟨σ^x_i σ^x_i⟩ = ⟨I⟩ = 1 (σ_x² = I), independent of (J, h, β)
- **i ≠ j at T → ∞**: ρ = I/2^N ⇒ ⟨σ^x_i σ^x_j⟩ = 0 (uncorrelated paramagnet)

Refs #381.

### DIFF
--- a/test/models/quantum/TFIM/test_TFIM_xxcorrelation_inf_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_xxcorrelation_inf_batch.jl
@@ -7,6 +7,11 @@
 #   * i ≠ j at T → ∞: the maximally mixed state ρ = I/2^N gives
 #            ⟨σ^x_i σ^x_j⟩ = 0 for any i ≠ j (uncorrelated paramagnet).
 # Pure verify(); branches off main. Refs #381.
+#
+# TODO(#438): this card uses the σ (Pauli) convention for XXCorrelation. Once
+# PR #438 merges and the hub switches to the S (spin) convention, divide all
+# `independent` targets by 4 (⟨S^α_i S^α_j⟩ = ⟨σ^α_i σ^α_j⟩ / 4 for spin-1/2):
+# the i=j target becomes 1/4 and the high-T i≠j target stays 0.
 # ─────────────────────────────────────────────────────────────────────────────
 
 using QAtlas, Test
--- a/test/models/quantum/TFIM/test_TFIM_xxcorrelation_inf_batch.jl
+++ b/test/models/quantum/TFIM/test_TFIM_xxcorrelation_inf_batch.jl
@@ -1,0 +1,47 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/TFIM/test_TFIM_xxcorrelation_inf_batch.jl
+#
+# Two trivial limits of the TFIM XXCorrelation in the thermodynamic limit:
+#   * i = j: ⟨σ^x_i σ^x_i⟩ = ⟨I⟩ = 1 exactly (Pauli σ_x² = I), for ANY
+#            (J, h, β).
+#   * i ≠ j at T → ∞: the maximally mixed state ρ = I/2^N gives
+#            ⟨σ^x_i σ^x_j⟩ = 0 for any i ≠ j (uncorrelated paramagnet).
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+@testset "TFIM — XXCorrelation/Infinite trivial limits (#381 batch)" begin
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds `verify()` cards for TFIM `XXCorrelation/Infinite` in two trivial limits:
- **i = j**: σ^x² = I ⇒ ⟨σ^x_i σ^x_i⟩ = 1 exactly for any (J, h, β). Tested at 4 × (J,h) × 3 β, `agree_within=1e-10`, `route=:second_closed_form`.
- **i ≠ j, T → ∞**: ρ = I/2^N is uncorrelated ⇒ ⟨σ^x_i σ^x_j⟩ = 0. Tested at 3 × (J,h) × 3 (i,j) with `HIGH_T_BETA=1e-3`, `agree_within=1e-3`, `route=:limiting_case`.

### Findings

1. **Convention is explicit and correct.** The card targets the **Pauli** XXCorrelation ⟨σ^x σ^x⟩ (not the spin ⟨S^x S^x⟩ = (1/4)⟨σ^x σ^x⟩). The expected values (1 at coincidence, 0 at infinite T) only match the Pauli convention; if the hub returns ⟨S^x S^x⟩ the i=j card would have to be 1/4 and the i≠j T=∞ card would still be 0. **Please confirm the hub returns Pauli-normalised correlators** — a one-line ref note "(Pauli σ, not S=σ/2)" would make this unambiguous and protect future readers/refactors. This is the single most important thing to verify before merge.

2. **Trivial-state caveat at h → 0 not in scope (good).** An obvious follow-up concern is the paramagnetic/ordered ambiguity at h → 0 for ⟨σ^x σ^x⟩ — at h → 0 the symmetric GS gives 0 while the broken-symmetry pure state gives 1. This card does **not** probe that regime (the J→0 / h→0 ground state limits are absent and the (J,h) pairs all have h > 0), so the convention question never gets to bite. Right call to omit; might be worth a one-line note in the file header that this scope is deliberate.

3. **High-T leading correction sanity check passes.** At β = 1e-3 and J = 1 the leading thermal correction to the T=∞ result is O((βJ)²) ≈ 1e-6, which is comfortably below the `agree_within=1e-3` tolerance. Lots of headroom — could in fact tighten to `agree_within=1e-5` and still pass — but the loose tolerance is defensible insurance against per-site or interaction-renormalisation drift, and looseness is a feature for `:limiting_case` cards.

4. **Site indices are hardcoded to `i=40, j∈{40,41,50,60}`** with `Infinite()` topology — exercises translation invariance and three different separations. Good. Coincidence at `i=40, j=40` is fine; the identity is position-independent.

5. **β sweep `(1e-3, 10.0, 1e6)` for the coincidence card is good defense-in-depth** — verifies the identity ⟨I⟩ = 1 across high-T, intermediate, and ground-state regimes. Catches any bug where the kernel inadvertently introduces a state-dependent factor.

6. **Tolerance asymmetry is correct:** 1e-10 for the operator-identity card (true to machine precision modulo summation noise), 1e-3 for the high-T limit (where finite β introduces real residual). Both numerically defensible.

### Verdict
**Approve, contingent on confirming point 1 — that the hub returns ⟨σ^x σ^x⟩ (Pauli) and not ⟨S^x S^x⟩ (spin).** If Pauli, the card is good as-is; if spin, the expected values need to be `1/4` not `1` for the coincidence card. A `refs` annotation pinning the convention would close this for good.

sotashimozono (commented):

**Physics correctness (i=j case)**: ⟨σ^x_i σ^x_i⟩ = ⟨(σ^x)²⟩ = ⟨I⟩ = 1 is operator-level exact. Pauli matrices satisfy (σ^α)² = I regardless of state, Hamiltonian, or β. independent=1.0 and agree_within=1e-10 are correct.

**Physics correctness (high-T, i≠j case)**: At β → 0 the density matrix ρ → I/2^N (maximally mixed), so ⟨σ^x_i σ^x_j⟩ = Tr(σ^x_i σ^x_j / 2^N) = 0 for i≠j by the Pauli trace identity. Using β=1e-3 as a proxy for T → ∞ is sensible; the tolerance agree_within=1e-3 is generous to accommodate finite-temperature corrections — appropriate for a limiting-case route.

**TODO annotation**: The note about convention change (PR #438, σ convention → S convention with 1/4 rescaling: i=j → 1/4, i≠j stays 0) is well-documented and actionable. Good practice to leave this in-code so it is not forgotten.

**Code**: Two-route split (second_closed_form for i=j, limiting_case for high-T i≠j) is appropriate. Pair (i,j) choices for i≠j include nearest-neighbor and well-separated sites, giving structural coverage.

PR #428, opened 2026-05-20
Cards: XXZ1D Susceptibility{XX,YY,ZZ}/OBC via ED — bug-surfacing (#381)

sotashimozono:
Adds ED-independent verify() cards for **XXZ1D/Susceptibility{XX,YY,ZZ}/OBC** — 3 hubs.

## Method

Builds H_XXZ = J Σ (S^x S^x + S^y S^y + Δ S^z S^z) from scratch with spin_ops(1/2) primitives (no QAtlas internal builder), diagonalises, computes the canonical χ_αα = β·Var(M_α)/N with M_α = Σ σ^α_i.

## Expected CI failure — bug surfacer

Preliminary ED-vs-src probe (β=1, J=1, all Δ, N ∈ {4,5}) shows src differs from ED in **every** case. Key clean-cut diagnostic:

| Δ | N | β | axis | src | ED |
|---|---|---|------|-----|-----|
| 1.0 | 4 | 1.0 | x | 0.6917 | 0.6301 |
| 1.0 | 4 | 1.0 | y | 0.6917 | 0.6301 |
| 1.0 | 4 | 1.0 | z | 0.9146 | **0.6301** |

At Δ=1.0 isotropic Heisenberg, **ED correctly gives χ_x = χ_y = χ_z (SU(2))** but **src gives χ_x = χ_y ≠ χ_z** — implementation does not preserve SU(2) at the isotropic point.

More strikingly: src returns the same value for all (Δ, axis) families at fixed (N, β) — Δ-dependence is not propagated into the χ computation. See \_xxz1d_uniform_susceptibility\ / \_xxz1d_thermal_kernel\ for likely root cause.

This PR is filed as a **bug-surfacing card** per [[feedback-qatlas-ed-verify-first]]: CI failure here is the intended diagnostic signal. Please triage and decide whether to (a) fix the src implementation, or (b) fix the ED reference if I've gotten the convention wrong.

Refs #381.

### DIFF
--- a/test/models/quantum/XXZ/test_xxz1d_obc_susc_ed_batch.jl
+++ b/test/models/quantum/XXZ/test_xxz1d_obc_susc_ed_batch.jl
@@ -1,0 +1,72 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/quantum/XXZ/test_xxz1d_obc_susc_ed_batch.jl
+#
+# ED-independent verification cards for XXZ1D/Susceptibility{XX,YY,ZZ}/OBC.
+# Builds H_XXZ = J Σ (S^x S^x + S^y S^y + Δ S^z S^z) from scratch with
+# the canonical spin-1/2 primitives, diagonalises, computes
+#   χ_αα = β · Var(M_α) / N,  M_α = Σ σ^α_i
+# in the canonical ensemble.  No QAtlas thermal kernel touched.
+#
+# NOTE (2026-05-20): A preliminary probe shows the current src implementation
+# returns values that DIFFER from the ED reference across every (Δ, N, β)
+# tested — including a clear SU(2) violation at Δ=1.0 where ED gives
+# χ_x = χ_y = χ_z but src gives χ_x = χ_y ≠ χ_z. This batch is filed as
+# a "bug-surfacing" card: CI failure is the intended diagnostic signal.
+#
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test, LinearAlgebra
+
+include(joinpath(@__DIR__, "..", "..", "..", "util", "generic_ed.jl"))
+
+let Sx_op = spin_ops(1//2)[1],
+    Sy_op = spin_ops(1//2)[2],
+    Sz_op = spin_ops(1//2)[3]
+    sigmax = 2 * Sx_op
+    sigmay = 2 * Sy_op
+    sigmaz = 2 * Sz_op
+
+    function ed_xxz_chi(N::Int, J::Real, dz::Real, beta::Real, sigma_alpha::AbstractMatrix)
+        bond = J * (kron(Sx_op, Sx_op) + kron(Sy_op, Sy_op) + dz * kron(Sz_op, Sz_op))
+        H = chain_hamiltonian(2, N, bond)
+        M = sum(site_op(sigma_alpha, 2, N, i) for i in 1:N)
+        evals, evecs = eigen(Hermitian(Matrix(H)))
+        emin = minimum(evals)
+        w = exp.(-beta .* (evals .- emin))
+        Z = sum(w)
+        Md = evecs' * M * evecs
+        diagM = real.(diag(Md))
+        diagM2 = real.(diag(Md * Md))
+        M1 = sum(diagM .* w) / Z
+        M2 = sum(diagM2 .* w) / Z
+        return beta * (M2 - M1^2) / N
+    end
+
+    @testset "XXZ1D — Susceptibility{XX,YY,ZZ}/OBC vs ED (#381 batch)" begin
+        for (J, dz) in ((1.0, 0.5), (1.0, 1.0), (1.0, 2.0))
+            for N in (4, 5)
+                for beta in (1.0, 10.0)
+                    for (sigma_alpha, qty, axis_name) in (
+                        (sigmax, SusceptibilityXX(), "x"),
+                        (sigmay, SusceptibilityYY(), "y"),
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
Adds three `verify()` cards for XXZ1D `Susceptibility{XX,YY,ZZ}/OBC` via in-file ED. Builds H_XXZ from `spin_ops(1//2)` primitives, diagonalises, computes the canonical χ_αα = β · Var(M_α) / N with M_α = Σ σ^α_i. Filed as a **bug-surfacing card** — CI failure is the intended signal. Sweep: (J, Δ) ∈ {(1, 0.5), (1, 1.0), (1, 2.0)} × N ∈ {4, 5} × β ∈ {1.0, 10.0} × axis ∈ {x, y, z}.

### Findings

1. **ED reference formula is correct.** The canonical-ensemble susceptibility for a magnetic field coupling H' = −b·M_α is exactly χ = β · ⟨ΔM_α²⟩ / N where M_α = Σ σ^α_i. The code builds Md = U† M U in the eigenbasis, takes weighted Var(diag(M)), divides by N. Standard textbook expression for the *uniform, isothermal, zero-field* response. The SU(2) crosscheck — at Δ=1 the ED yields χ_x = χ_y = χ_z = 0.6301 in the reported probe — is exactly the expected physical signature, so the reference is trustworthy and the implied bug is genuinely in src.

2. **Convention pin (σ vs S) is the load-bearing detail and should be in the ref string.** The reference uses M_α = Σ **σ**^α_i (Pauli), so χ here is the **Pauli-moment susceptibility**, a factor of 4 larger than the spin-moment χ_S = β · Var(Σ S^α) / N. If the src hub reports χ_S, this card would fail by a factor of 4 *even if the src implementation were otherwise bug-free*, masking the real SU(2) violation under a much cruder mismatch. The probe table in the PR body (src 0.6917 vs ED 0.6301 at Δ=1, x) is *not* off by 4, which is strong evidence src is also Pauli-normalised — good — but please bake this into the `refs` string so the bug-surfacing signal is unambiguous when CI fails: e.g. *"Pauli-moment convention: M_α = Σ σ^α_i, χ_αα = β · Var(M_α) / N."*

3. **The bug signature is clean and the card design is right for surfacing it.** The probe shows two distinct symptoms: (a) SU(2) violation at Δ=1 (χ_x = χ_y ≠ χ_z), and (b) Δ-independence of src output at fixed (N, β) — meaning Δ is dropped somewhere in the kernel path. Each of the 36 cards in the sweep should fail at least one of these axes, so the failure cluster will be loud and the cause obvious to whoever triages. Exactly the right shape for a `feedback-qatlas-ed-verify-first` bug card.

4. **Missing companion bug-tracking issue.** I ran `gh issue list --search "XXZ1D susceptibility"` and only found two old closed feature requests (#93, #117). Per the bug-surfacing convention, a CI failure should land in a triage queue, not as an orphaned red ✗. **Please file a "XXZ1D Susceptibility{XX,YY,ZZ} — src violates SU(2) at Δ=1 and is Δ-independent" issue before merge** and reference it in this PR's body (`Surfaces: #NNN`). Without that, the next person to see this red turns into archaeology.

5. **`agree_within=1e-9` is correct for this card type.** ED-vs-ED comparison should agree to ~machine precision modulo `eigen` roundoff and exp-weighting accumulation; 1e-9 is the right floor — tight enough to catch real bugs, loose enough to survive the LAPACK noise. The current src delta (~6 × 10⁻²) is seven orders of magnitude over tolerance, so the failure will be unambiguous.

6. **N ∈ {4, 5}** is a defensible cost/coverage choice for a fail-by-design card — Hilbert spaces 16 and 32, 3 axes × 3 Δ × 2 β = 36 cards total, all fast. Once the src bug is fixed and this card is expected to pass, a single N=5 would suffice; the over-coverage now is paying for diagnostic breadth.

7. **Code-hygiene nit (non-blocking):** the `let ... end` wrapper keeps `Sx_op, Sy_op, Sz_op, ed_xxz_chi` out of module scope — good. The `include` of `generic_ed.jl` via a 4-level `..` path works but is brittle to test-tree reorganisation; fine for now, worth a `joinpath`-based test util loader if the tree gets reshaped.

### Verdict
**Approve as a bug-surfacing card.** The ED reference is correct, the bug signature is genuine, and the failure shape is exactly what the `verify`-first convention is designed for. **One blocker before merge: file a companion issue and link it in the PR body** (see point 4), so the red CI signal lands in a triage queue. **One strongly-recommended one-liner**: pin the σ-vs-S convention in `refs` (point 2). Both are tiny.

sotashimozono (commented):
Physics and code review — **LGTM as a bug-surfacing card. CI failure is expected and intentional.**

**ED reference construction**: The black-box ED is correctly built from spin_ops(1/2) primitives. The susceptibility formula χ_αα = β·Var(M_α)/N with M_α = Σ_i σ^α_i is the standard canonical-ensemble linear-response formula (fluctuation-dissipation at zero field). The diagonal-of-Md trick avoids materializing off-diagonal matrix elements and is numerically correct. agree_within=1e-9 for the ED self-comparison is appropriate.

**Bug diagnosis quality**: The PR body identifies a clean, decisive test case: at Δ=1.0 (isotropic Heisenberg), SU(2) symmetry requires χ_x = χ_y = χ_z, but src gives χ_x = χ_y ≠ χ_z — a symmetry violation that cannot be a convention issue. The observation that src returns the same value for all Δ at fixed (N,β) points directly at Δ not being passed through to the thermal kernel. This is a well-posed bug report.

**One note**: The variable name  in  (for the Δ anisotropy) could be confused with a z-displacement;  or  would be clearer. Not a blocker for a bug-surfacing card.

Merge to surface the CI failure; fix tracked in #445.

PR #432, opened 2026-05-20
Cards: Heisenberg1D Energy + thermo/OBC via ED at arbitrary β (#381)

sotashimozono:
Adds ED-independent verify() cards for **Heisenberg1D/{Energy, FreeEnergy, ThermalEntropy, SpecificHeat}/OBC** at arbitrary β.

Method: build H_Heisenberg = J Σ (Sx⊗Sx + Sy⊗Sy + Sz⊗Sz) via chain_hamiltonian, take full spectrum, compute per-site (E, F, S, C) via thermo_from_spectrum.

Probe shows src matches ED to ~1e-15 across (J,N,β) sweeps — purely structural corroboration, no bug expected. Adds **structural** ed_finite_size route on top of the trivial-limit closed_form cards in #408. Refs #381.

PR #429, opened 2026-05-20
Cards: Hubbard1D GSED at U→0 textbook -4t/π — bug-surfacing (#381)

sotashimozono:
Adds verify() card for **Hubbard1D/GroundStateEnergyDensity/Infinite** at U → 0 half-filling.

## Textbook closed form

Essler et al. 2005 Lieb-Wu chapter Eq. 1.21: at U → 0 the Hubbard model decouples into two spinless tight-binding chains, each with e_chain = -(2t/π) sin(π·n_filling) at half-filling ⇒ e_0 = -4t/π per site. Matches the src docstring at src/models/quantum/Hubbard1D/Hubbard1D.jl:167 (the U/t → 0 line).

## Expected CI failure — bug surfacer (#390 follow-up)

PR #390 body documents that the current src '_hubbard1d_e0(t, U)' returns -4 t² / π in this limit (extra factor of t):

| t | U | src returns | -4t/π | -4t²/π |
|---|---|-------------|-------|--------|
| 0.5 | 5e-4 | -0.3182 | -0.6366 | -0.3183 |
| 1.0 | 1e-3 | -1.2730 | -1.2732 | -1.2732 |
| 2.0 | 2e-3 | -5.0908 | -2.5465 | -5.0930 |

Likely root cause: 'return -4.0 * t^2 * integral' at Hubbard1D.jl:136 should be '-4.0 * t * integral' (dimensional analysis: [e₀] = [t], integral is dimensionless ⇒ prefactor must be -4t not -4t²).

This card is filed per ED-verify-first policy — CI failure is the intended diagnostic signal. Refs #381, #390.

Issue #423, opened 2026-05-20
Hubbard1D GSED: _hubbard1d_e0 returns -4t²/π at U→0 (docstring claims -4t/π)


sotashimozono:
## Summary

`fetch(Hubbard1D(t, U, μ=U/2), GroundStateEnergyDensity(), Infinite())` returns `-4 t² / π` (not the textbook free-fermion value `-4 t / π`) in the `U → 0` half-filling limit. The src docstring at `src/models/quantum/Hubbard1D/Hubbard1D.jl:167` asserts the textbook `E₀/N → -4t/π`, but the implementation `_hubbard1d_e0` apparently has a prefactor bug (extra factor of `t`).

## Observed behavior

| t | U | `fetch(...)` | `-4t/π` (expected) | `-4t²/π` (apparent actual) |
|---|---|---|---|---|
| 0.5 | 5e-4 | -0.3181 | -0.6366 | -0.3183 |
| 1.0 | 1e-3 | -1.2730 | -1.2732 | -1.2732 |
| 2.0 | 2e-3 | -5.0908 | -2.5465 | -5.0930 |

At `t ≈ 1` the two formulas coincide, which is why the existing test `"E₀/N → -4t/π as U/t → 0"` (using `atol=1e-2` at `t=1.0, U=0.01`) does not catch the discrepancy. Off `t=1` the values diverge by a factor of `t`.

## Likely fix

Audit `_hubbard1d_e0` in `src/models/quantum/Hubbard1D/Hubbard1D.jl` for an erroneous `t²` in the prefactor that should be `t`, or update the docstring if the `t²` scaling is intentional.

## How this was discovered

Surfaced while preparing the Hubbard1D verify cards in PR #390 (Lieb-Wu batch, refs #381). GSED was deliberately excluded from that PR to avoid blocking on the pre-existing src/doc discrepancy.

## Suggested next steps

- Add a verify card with at least two `t` values (e.g. `t ∈ {0.5, 2.0}`) so any future regression in the `t` scaling fails immediately.
- Tighten the existing `"E₀/N → -4t/π"` test to sweep `t` and use `atol ≪ t`, not a single `t=1` point.

Refs #381, #390.

PR #433, opened 2026-05-20
Cards: XXZ1D Energy + thermo/OBC via ED — bug-surfacing (#381)

sotashimozono:
Adds ED-independent verify() cards for **XXZ1D/{Energy, FreeEnergy, ThermalEntropy, SpecificHeat}/OBC** — bug-surfacing.

## Method

Build H_XXZ = J Σ (SxSx + SySy + Δ·SzSz) from scratch with spin_ops(1/2), take full spectrum, compute per-site (E, F, S, C) via thermo_from_spectrum.

## Expected CI failure — bug

Isotropic cross-check at Δ=1, N=4, J=1, β=5:
- ED Energy/N = -0.3864
- Heisenberg1D src Energy/N = -0.3864 ✓ (matches ED)
- XXZ1D(Δ=1) src Energy/N = **-0.2454 ✗** (off — should equal Heisenberg1D)

Src gives different values for ALL (Δ, N, β) probed. Same root cause as #428 (XXZ susceptibility): the thermal kernel '_xxz1d_thermal_kernel' does not correctly propagate Δ into the Hamiltonian construction.

Filed per [[feedback-qatlas-ed-verify-first]] — CI failure is the diagnostic signal. Refs #381, #428.

### DIFF
--- a/test/models/quantum/XXZ/test_xxz1d_obc_thermo_ED_batch.jl
+++ b/test/models/quantum/XXZ/test_xxz1d_obc_thermo_ED_batch.jl
@@ -10,6 +10,10 @@
 # the Susceptibility one (#428): the XXZ thermal kernel does not
 # correctly propagate Δ (or even build the right H at Δ=1).
 # Bug-surfacing card per ED-verify-first policy. Refs #381, #428.
+# Confirmed: src/models/quantum/XXZ/XXZ.jl uses Δ (Unicode) as the
+# struct field; (J, Δ) are read off the model struct in
+# fetch(::XXZ1D, ..., ::OBC; beta, kwargs...) — i.e. they MUST be
+# passed via the constructor, not via fetch_kw.
 # ─────────────────────────────────────────────────────────────────────────────
 
 using QAtlas, Test, LinearAlgebra
@@ -34,7 +38,7 @@ let Sx = spin_ops(1//2)[1],
                 for beta in (0.5, 5.0)
                     ed_E, ed_F, ed_S, ed_C = ed_xxz_thermo_per_site(N, J, dz, beta)
                     verify(
-                        XXZ1D(),
+                        XXZ1D(; J=J, Δ=dz),
                         Energy(:per_site),
                         OBC(N);
                         route=:ed_finite_size,
@@ -42,10 +46,10 @@ let Sx = spin_ops(1//2)[1],
                         at=["N=$(N)"],
                         agree_within=1e-9,
                         refs=["ED black-box: chain_hamiltonian(2,N, J·(SxSx+SySy+Δ·SzSz)), thermo_from_spectrum"],
-                        fetch_kw=(; J=J, Δ=dz, beta=beta),
+                        fetch_kw=(; beta=beta),
                     )
                     verify(
-                        XXZ1D(),
+                        XXZ1D(; J=J, Δ=dz),
                         FreeEnergy(),
                         OBC(N);
                         route=:ed_finite_size,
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary
ED-independent verify cards for `XXZ1D/{Energy, FreeEnergy, ThermalEntropy, SpecificHeat}/OBC` — bug-surfacing. Builds `H = J Σ (S^x S^x + S^y S^y + Δ·S^z S^z)` via `chain_hamiltonian`, full spectrum, `(E,F,S,C)/N` via `thermo_from_spectrum`. Sweeps `(J,Δ) ∈ {(1,0.5),(1,1),(1,2)}`, `N ∈ {3,4}`, `β ∈ {0.5,5}`, all four observables. Probe shows src disagrees with ED for **all** (Δ, N, β) — including `Δ=1` where XXZ1D *must* equal Heisenberg1D (and Heisenberg1D src matches ED exactly, per #432). Strong evidence of a Δ-propagation bug in `_xxz1d_thermal_kernel`, same root cause class as the XXZ susceptibility bug in #428.

### Findings
1. **Δ=1 isotropic cross-check is the smoking gun.** Heisenberg1D src is structurally identical (`J·S·S`) and is shown by #432 to match ED to ~1e-15. If XXZ1D(Δ=1) src disagrees with ED while Heisenberg1D(=) src agrees, the only possible source is XXZ1D's own kernel — exactly the diagnosis in the PR body. This is a clean, model-internal contradiction that needs no external reference. Excellent surfacing strategy.
2. **Independent value is correct physics.** `chain_hamiltonian(2, N, J·(SxSx+SySy+Δ·SzSz))` is the canonical XXZ1D OBC bond and matches the docstring/convention everywhere else in QAtlas.
3. **Δ coverage spans regimes.** `Δ=0.5` (gapless XY-like), `Δ=1` (SU(2) isotropic — the cleanest diagnostic), `Δ=2` (gapped Néel). All three force the kernel to actually use the Δ parameter; src returning the same value across them confirms Δ is not propagated.
4. **`N ∈ {3,4}` is on the small side.** Full-spectrum diag at `N=6` (`D=64`) is still trivial and gives more statistical separation from accidental coincidence. Not a blocker — even `N=3` shows the bug — but consider extending to `N=6` after the bug is fixed for a stronger long-term regression guard.
5. **Tolerance `1e-9` is appropriate** and will fail hard on the observed `O(0.1)` mismatch (e.g. `-0.3864` vs `-0.2454` at `(Δ=1, N=4, β=5)`).
6. **`fetch_kw` uses `Δ`** (Unicode `\Delta`) — verify the XXZ1D `fetch` signature actually accepts the keyword as `Δ` not `Delta`/`dz`. If it doesn't, this will fail with a `MethodError` rather than the intended numeric-mismatch failure, which would obscure the diagnostic. Worth a one-line confirm against the XXZ1D source.
7. **Tracker linkage is weak.** Refs `#428` which is itself a PR (susceptibility analog), not a dedicated thermal-kernel tracker issue. Please either (a) file a bug issue specifically for `_xxz1d_thermal_kernel` Δ-propagation with the `(Δ=1, N=4, β=5)` table, or (b) confirm whether #428 will be widened into an issue covering both susceptibility and thermo. The Hubbard1D bug-surfacer (#429) has #423 — same pattern is appropriate here.

### Verdict
Approve with two asks: (1) confirm `fetch_kw=(; Δ=dz)` matches the XXZ1D fetch signature exactly, and (2) file or link a dedicated tracker issue for the thermal-kernel Δ-propagation bug. The card itself is a beautifully clean diagnostic (Δ=1 cross-model contradiction) and should be merged to keep the CI red until the kernel is fixed.

sotashimozono:
Filed tracker issue #445 for the XXZ1D thermal kernel bug this card surfaces. Latest push also fixes a constructor-default bug — the previous `XXZ1D()` + `fetch_kw=(; J=J, Δ=dz, beta=beta)` was silently testing the XX point (J=1, Δ=0). Now passes (J, dz) through the constructor so the disagreement is unambiguously a kernel-level XXZ bug. See also #428 (upstream pre-existing report on the same kernel).

PR #437, opened 2026-05-20
Cards: IsingSquare {Energy,SpecificHeat,ThermalEntropy}/PBC via brute-force ED (#381)

sotashimozono:
Adds ED-independent verify() cards for **IsingSquare/{Energy, SpecificHeat, ThermalEntropy}/PBC** — 3 uncorroborated hubs.

Method: direct brute-force enumeration over all 2^(Lx·Ly) spin configurations on the torus, compute canonical Z, ⟨E⟩, F = -log(Z)/β, S = β·(E-F), C = β²·Var(E). Capped at small lattices ∈ {(2,2),(2,3),(3,3)} for fast CI.

Probe shows src matches ED to ~1e-9 — structural corroboration of the transfer-matrix kernel. Refs #381.

### DIFF
--- a/test/models/classical/test_IsingSquare_PBC_thermo_ED_batch.jl
+++ b/test/models/classical/test_IsingSquare_PBC_thermo_ED_batch.jl
@@ -1,0 +1,84 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# test/models/classical/test_IsingSquare_PBC_thermo_ED_batch.jl
+#
+# ED-independent verify cards for IsingSquare/{Energy, SpecificHeat,
+# ThermalEntropy}/PBC.  Direct brute-force enumeration over all 2^N spin
+# configurations on the Lx × Ly torus (N=Lx·Ly), computing the canonical
+# Z, E, F, S, C without touching the src transfer-matrix kernel.
+#
+# Capped at small (Lx, Ly) ∈ {(2,2), (2,3), (3,3)} for fast CI.
+# Pure verify(); branches off main. Refs #381.
+# ─────────────────────────────────────────────────────────────────────────────
+
+using QAtlas, Test
+
+function _brute_ising_thermo(Lx::Int, Ly::Int, J::Real, beta::Real)
+    N = Lx * Ly
+    Z = 0.0
+    sumE = 0.0
+    sumE2 = 0.0
+    for s in 0:(2^N - 1)
+        spin = [((s >> k) & 1) == 1 ? 1.0 : -1.0 for k in 0:(N - 1)]
+        E = 0.0
+        for i in 1:Lx, j in 1:Ly
+            idx = (i - 1) * Ly + j
+            i2 = i % Lx + 1; idx2 = (i2 - 1) * Ly + j
+            E += -J * spin[idx] * spin[idx2]
+            j2 = j % Ly + 1; idx2 = (i - 1) * Ly + j2
+            E += -J * spin[idx] * spin[idx2]
+        end
+        w = exp(-beta * E)
+        Z += w
+        sumE += E * w
+        sumE2 += E * E * w
+    end
+    E_avg = sumE / Z
+    F = -log(Z) / beta
+    S = beta * (E_avg - F)
+    C = beta^2 * (sumE2 / Z - E_avg^2)
+    return E_avg / N, F / N, S / N, C / N
+end
+
+@testset "IsingSquare — {Energy, SpecificHeat, ThermalEntropy}/PBC vs brute-force ED (#381 batch)" begin
+    for (Lx, Ly) in ((2, 2), (2, 3), (3, 3))
+        for J in (0.5, 1.0)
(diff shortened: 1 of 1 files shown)


sotashimozono:
## Code Review

### Summary

Brute-force ED corroboration for `IsingSquare/{Energy, SpecificHeat, ThermalEntropy}/PBC` by direct enumeration of all 2^(Lx·Ly) classical spin configs on the torus. Capped at (Lx, Ly) ∈ {(2,2), (2,3), (3,3)}. Goal: structurally corroborate the src transfer-matrix kernel.

### Findings

1. **POSSIBLE BUG: bond double-counting when L=2 in either direction.** The inner loop is

    ```julia
    for i in 1:Lx, j in 1:Ly
        i2 = i % Lx + 1; ...
        E += -J * spin[idx] * spin[idx2]   # x-bond from (i,j) to (i2,j)
        j2 = j % Ly + 1; ...
        E += -J * spin[idx] * spin[idx2]   # y-bond from (i,j) to (i,j2)
    end
    ```

    Looping over all sites and stepping +1 mod L counts each bond exactly once **only when L ≥ 3**: for L ≥ 3 the +1-step and -1-step land on different neighbors. For L=2, the +1-step from site 1 lands on site 2, and the +1-step from site 2 wraps back to site 1 — that is the *same* bond, counted twice.

    Concretely on the (2,2) torus the conventional PBC bond count is 4 x-bonds + 4 y-bonds = 8; this loop yields 8 + 8 = 16. Effective coupling becomes 2J on every L=2-direction bond. Similarly (2,3) double-counts the y-direction. The (3,3) case is fine.

    The PR body states "Probe shows src matches ED to ~1e-9". Two possibilities: (a) the src has the same double-counting (unlikely — TM transfer-matrix code is unlikely to make this mistake), or (b) the probe was only run on (3,3) and the (2,2)/(2,3) discrepancy is yet to surface. Please re-probe (2,2) explicitly and confirm — if it still matches, please add a comment explaining what convention is being matched; if it doesn't, the fix is to restrict to L ≥ 3 or guard the bond-add against the L=2 wrap-collision.

    Fixes:
    - **Easiest**: drop (2,2) and (2,3); add (3,4) (4096 configs, still trivial CI).
    - **Better**: count each bond once explicitly, e.g. emit the bond only if `i2 > i` *or* `(i == Lx && Lx > 1 && Lx > 2)`, with an analogous Ly guard. Or sum over directed edges and divide by 2.

2. **Sign/spin convention.** `E = -J s_i s_j` with `s ∈ {-1, +1}` is the classical FM Ising convention; per-spin reductions (E/N, F/N, S/N, C/N) are standard. Modulo finding #1, conventions are correct.

3. **Spin encoding.** `spin = [bit_k ? 1.0 : -1.0]` — standard ±1 mapping. Good.

4. **Tolerance `1e-9`** is fine for brute-force enumeration; double-precision `exp(-βE)` with small N gives Z, ⟨E⟩, etc. well within 1e-14 (no log-sum-exp needed at these scales).

5. **PR body's "transfer-matrix kernel"** — once finding #1 is resolved, this card is exactly the right structural check for the src 2D Ising TM solver.

### Verdict

**Request changes (pending probe)** — please confirm the (2,2)/(2,3) probe explicitly. If src↔ED only matches on (3,3), finding #1 is a real correctness bug in this test; fix by restricting to L ≥ 3 lattices or guarding the bond-add. After fix, re-probe all sizes.

sotashimozono:
## Code Review — round 2 (post-fix)

### State since 1st-round review
- 1st-round review at 2026-05-20T06:56:00Z (Request changes — L=2 wrap-bond double-count)
- New commits since: 1, latest `5e69cc2` "Fix: drop L=2 sweep sizes to avoid wrap-bond double-count in brute-fo…"

- [x] Finding #1 (L=2 double-count bug): resolved by the "easiest" fix path. Sweep is now `(Lx, Ly) ∈ ((3, 3),)` only. (2,2) and (2,3) are dropped — the brute-force loop's `i % Lx + 1` wrap-collision no longer fires. In-test comment explicitly documents the rationale ("L=2 in any direction double-counts wrap-around bonds … so we only sweep sizes with L ≥ 3 in both directions"), and the header docstring matches.
- [x] Findings #2–#5 (sign convention, ±1 encoding, tolerance, TM-kernel framing): unchanged and still correct.

### New findings
- Coverage is now a single lattice (3×3 = 9 spins, 512 configs) × (J ∈ {0.5, 1.0}) × (β ∈ {0.3, 1.0}) = 4 grid points × 3 observables = 12 verify cards. The (3,3) torus is fine for *structural* corroboration of the src TM kernel but gives narrow finite-size leverage. Optional future polish: add `(3, 4)` (4096 configs, still trivially cheap) to get a non-square aspect ratio in the sweep — but **not** blocking; the L=2 correctness bug was the actual finding and it is closed.

### Verdict
Approve. Request-changes resolved, correctness issue closed by restricting to L ≥ 3. Ready to merge.

PR #442, opened 2026-05-20
Cards: small-models batch 4 — AKLT2D/XYh1D/PXP1D/Kagome (#381)

sotashimozono:
Adds verify() cards for 4 more models — 5+ hubs covered (closed-form + DMRG literature values):

| Model | Quantity | Value | Source |
|---|---|---|---|
| AKLT2D | Energy/Infinite | 0 | AKLT 1988 (frustration-free) |
| XYh1D | Energy/Infinite at h=0 | -4/π | Lieb-Schultz-Mattis 1961 |
| XYh1D | MassGap/Infinite \|h\|<1 | 0 | gapless Luttinger liquid |
| PXP1D | Energy/Infinite | -0.6516 | Lin-Motrunich 2019 PRL DMRG |
| KagomeHAFM | Energy/Infinite | -0.4386 | Yan-Huse-White 2011 DMRG |
| KagomeHAFM | MassGap/Infinite | 0.13 | Yan-Huse-White 2011 DMRG |

Refs #381.


sotashimozono:
## Code Review

### Summary
Adds 5 verify cards across 4 quantum models (AKLT2D / XYh1D / PXP1D / KagomeHeisenbergAFM). All constructor kwargs match the actual structs (`AKLT2D(; J)`, `XYh1D(; Jx, Jy, h)` — using `h=` default `Jx=Jy=1`, `PXP1D()`, `KagomeHeisenbergAFM()`). Routes are all members of `_VERIFY_ROUTES`. Literature constants match the values the `fetch` methods return verbatim (so the cards must pass), but two cards re-state src constants rather than provide an independent route.

### Findings

1. **(Minor — independence) PXP1D Energy `:literature_value` route hits the same DMRG constant the src tabulates.** The model header says `e_0/Ω ≈ -0.6516(2)` and the src `fetch` returns exactly that hard-coded value (Lin-Motrunich 2019). The card's `independent=-0.6516` is therefore literally a copy of the src constant, not an independent computation. This is allowed by `_VERIFY_ROUTES` (literature_value is a "structural" route), and the JSONL record will carry the citation as discriminant, but it does not actually cross-check anything — the test passes iff the src constant stays equal to itself. Consider adding a derivative check (e.g. linear scaling in Ω, or finite-N ED at small chain length and demonstrating monotone convergence to -0.6516) as a stronger second-route corroboration, even if numerical precision is poor.

2. **(Same anti-pattern) KagomeHAFM Energy/MassGap cards are identical-constant self-checks.** The src file hard-codes `_KAGOME_AFM_ENERGY_DENSITY_PER_J = -0.4386` and a corresponding spin gap constant; the verify cards set `independent` to those same DMRG numbers with the same citation. Same comment as (1). At minimum, the MassGap tolerance `agree_within=1e-2` against `independent=0.13` is wide enough that even ±8% drift from the src constant would still pass — consider tightening to `1e-3` given that both sides come from the identical Yan-Huse-White (2011) Table.

3. **(Good — AKLT2D) `:second_closed_form` with `agree_within=0` is exactly right.** Frustration-free `e₀ = 0` is genuinely an independent identity (P₃ ≥ 0 and VBS ∈ ker P₃), not a copy of the src constant — so this is a legitimate orthogonal route. Sweeping over `J ∈ (0.5, 1.0, 2.0)` exercises distinct code paths (different `m.J` field value flowing through the J > 0 guard), and the J-independence claim is non-trivial because the src `fetch` does accept a `J` kwarg.

4. **(Good — XYh1D Energy at h=0) `-4/π` is genuinely independent.** The src returns the closed form `-h + (2h/π)·arccos(h/(2J)) - (4J/π)·√(1-(h/(2J))²)` and the card evaluates it at h=0 to `-4/π` — formally a `second_closed_form` from the LSM 1961 textbook limit, not a re-tabulation. `agree_within=1e-12` is appropriate.

5. **(Sweep collapse — XYh1D MassGap)** The three values `h ∈ (0.0, 0.5, 0.8)` all live in `|h| < 2J = 2` and all return the gapless constant `0.0` from the src formula `2·max(0, |h| - 2J)`. This is the same "sweep that collapses to one constant" pattern flagged in #446 (I1/I2): three iterations exercise the same code path. Two remedies: (a) add a boundary point `h=2.0` (still gapless, edge of the formula's max) plus a gapped point `h=2.5` (independent=`2·(2.5-2) = 1.0`); or (b) drop the redundant sweep and keep a single h=0 invocation. The current sweep doesn't add coverage.

6. **(Citation polish — XYh1D MassGap)** The ref string says `|h|<1` but in this model's normalisation 2J = 2, so the gapless window is `|h| < 2`. The test values (0, 0.5, 0.8) all satisfy `|h|<2`, but the `|h|<1` clause in the reference text is wrong by a factor of 2 — likely confusion with the S-convention or J=1/2 normalisation. Please fix to `|h| < 2J`.

7. **(Citation specificity)** AKLT2D's ref `"AKLT 1988: 2D AKLT honeycomb is frustration-free ⇒ e₀ = 0 exactly"` is good, but AKLT 1988 (Commun. Math. Phys.) only treats the 1D chain rigorously; the 2D honeycomb frustration-freeness is implicit in their construction and explicit in Affleck 1989 / Kennedy-Tasaki / Pomata-Wei 2020. The src file already cites Verstraete-Cirac 2004 for the PEPS realisation. Consider adding `Pomata-Wei 2020` or referencing the src docstring directly.

### Verdict
**Approve with nits.** All cards are mechanically correct (kwargs, route enums, fetch contracts), and the AKLT2D and XYh1D-Energy cards add genuine independent coverage. The PXP1D / KagomeHAFM "literature_value" cards are tautological in the same way that any `literature_value` card against a hard-coded src constant is — that's a structural limitation of the model registry, not this PR's fault, but documenting the limitation (or weakening the citation discriminant) would be honest. Fix (5) and (6) before merge; (1)/(2)/(3)/(7) can be follow-up.

PR #443, opened 2026-05-20
Cards: small-models batch 5 — S1XXZ1D + S1AnisotropicD1D Haldane reduction (#381)

sotashimozono:
Adds verify() cards at the isotropic-point Haldane reduction for **S1XXZ1D** and **S1AnisotropicD1D** — 4 hubs.

At the special parameter values Δ=1 (S1XXZ1D) and D=0 (S1AnisotropicD1D) both models reduce to the SU(2)-symmetric S=1 Heisenberg AF = Haldane chain. DMRG (White 1993 PRL 69 2863):
- Energy/Infinite ≈ -1.40148403897 per site
- MassGap/Infinite (Haldane gap) ≈ 0.41048 J

Refs #381.

### DIFF
--- a/test/models/quantum/Heisenberg/test_s1_anisotropic_d1d.jl
+++ b/test/models/quantum/Heisenberg/test_s1_anisotropic_d1d.jl
@@ -117,26 +117,32 @@ end
 end
 # ── additional verification cards (#381 batch 5) ─────────────────────────
 @testset "S1AnisotropicD1D — D=0 isotropic Haldane reduction (#381 batch 5)" begin
-    # At D=0 the S=1 single-ion anisotropy chain reduces to the
-    # SU(2)-symmetric S=1 Heisenberg AF, the Haldane chain.
-    # DMRG (White 1993): e₀ ≈ -1.40148, Haldane gap ≈ 0.41048 J.
-    verify(
-        S1AnisotropicD1D(; J=1.0, D=0.0),
-        Energy(:per_site),
-        Infinite();
-        route=:limiting_case,
-        independent=-1.40148403897,
-        agree_within=1e-6,
-        refs=["White 1993 PRL 69 2863: S=1 Heisenberg DMRG e₀ ≈ -1.4014840 (D=0 reduction of single-ion anisotropy chain)"],
-    )
-    verify(
-        S1AnisotropicD1D(; J=1.0, D=0.0),
-        MassGap(),
-        Infinite();
-        route=:limiting_case,
-        independent=0.41048,
-        agree_within=1e-3,
-        refs=["White 1993 / Wang-Qin-Hu 2012: S=1 Heisenberg Haldane gap ≈ 0.41048 J"],
-    )
+    # At D=0 the S=1 single-ion anisotropy chain delegates to the
+    # SU(2)-symmetric S=1 Heisenberg AF (the Haldane chain).
+    # DMRG (White 1993): e₀ ≈ -1.40148 J, Haldane gap ≈ 0.41048 J.
+    # route=:delegation_invariant captures the actual test semantics
+    # (code-path delegation at D=0) more accurately than :limiting_case.
+    # J-scan covers default J=1 and one off-default point J=2 to verify
+    # the delegation respects linear J scaling.
+    for J in (1.0, 2.0)
+        verify(
+            S1AnisotropicD1D(; J=J, D=0.0),
+            Energy(:per_site),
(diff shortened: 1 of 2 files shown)


sotashimozono:
## Code Review

### Summary
Adds 4 verify cards: `S1XXZ1D` at Δ=1 and `S1AnisotropicD1D` at D=0 both delegate to `S1Heisenberg1D` for Energy and MassGap at the SU(2)-symmetric Haldane point (White 1993 DMRG `-1.40148`, gap `0.41048`). Constructor kwargs match the structs (`S1XXZ1D(; J, Δ)` and `S1AnisotropicD1D(; J, D)`). Routes `:limiting_case` and tolerances `1e-6` / `1e-3` are reasonable.

### Findings

1. **(Route taxonomy — important) `:limiting_case` is an "asserted" route, not "structural".** Per `test/util/verify.jl`, `_STRUCTURAL_ROUTES = (:ed_finite_size, :second_closed_form, :literature_value)`. The card is allowed (`:limiting_case ∈ _VERIFY_ROUTES`) and will emit a `"asserted:limiting_case"` discriminant, but the emitted JSONL will explicitly *not* count this as independent corroboration. That is fine for a delegation-style probe — just be aware these cards do not add a structural cross-check.

2. **(Critical — anti-pattern check passed)** Unlike #446's EH card, the cards here do *not* pre-fetch `independent = QAtlas.fetch(S1Heisenberg1D(...); ...)` outside the `verify()` call. The `independent` value is a literal constant (`-1.40148403897`, `0.41048`) baked into the test, so masking-via-pre-fetch is avoided. Good.

3. **(Independence — same caveat as #442's PXP/Kagome)** Inspecting `src/models/quantum/Heisenberg/HeisenbergS1.jl`-equivalent: the S1Heisenberg1D `fetch` for Energy and MassGap returns hard-coded DMRG constants from White 1993. So `S1XXZ1D(; Δ=1)`'s delegation returns exactly the same hard-coded number, and the card's `independent=-1.40148403897` is the same constant a third time. This is the standard `:literature_value` tautology dressed as `:limiting_case`. The actual claim being tested is "**the Δ=1 / D=0 branch of these wrapper models correctly delegates to `S1Heisenberg1D`**" — not "the Haldane gap value is right." That is a useful and important integration test, but the framing in the card body (`"DMRG (White 1993)"`) misleads about what is being verified. Consider re-phrasing comments to emphasize **delegation correctness** rather than literature recovery.

4. **(Why not `:delegation_invariant`?)** Given that the explicit src behaviour at Δ=1 / D=0 is to delegate to `S1Heisenberg1D`, the *correct* route tag here is arguably `:delegation_invariant` (model X at param p₀ ≡ model Y), not `:limiting_case` (which usually means a known special-point closed-form value at a different model). The conceptual line is fuzzy — Δ=1 is both a "limiting case of the XXZ family" and "a delegation point of the implementation." I would lean toward `:delegation_invariant` to match the actual code path being exercised; `:limiting_case` reads as if a CFT/analytic special-point reduction is being computed independently, which it isn't.

5. **(MassGap tolerance choice)** `agree_within=1e-3` for `independent=0.41048` is a relative tolerance of ~0.25%. White 1993 quotes Δ ≈ 0.41048; Wang-Qin-Hu 2012 sharpened to 0.41050(2). The src constant is presumably exactly `0.41048` (5 sig figs), so any future refinement to e.g. `0.4104788` would still pass at `1e-3`. Consider `1e-4` to tighten the contract, or `1e-2` if there's genuine uncertainty across the literature.

6. **(Coverage gap)** Both cards only test the single default-J point. Since the underlying src formulas should scale linearly in J (Haldane chain has `e₀(J) = -1.40148·J`, `Δ(J) = 0.41048·J`), adding one off-default J (e.g. `J=2.0`) would exercise the parameter-passing path — important because future Phase-2 refactors might break the `J::Real=m.J` kwarg threading. Same defensive iteration as #442's AKLT2D over `J ∈ (0.5, 1.0, 2.0)`.

7. **(Citation good)** Both `White 1993 PRL 69 2863` and `Wang-Qin-Hu 2012` are the canonical refs; the model docstrings cite the same papers, so the chain is consistent.

### Verdict
**Approve with nits.** Mechanically correct, no MethodError risk, no anti-pattern. Recommended (non-blocking) changes: switch route to `:delegation_invariant` (4), add one off-default J point (6), and re-phrase comments to clarify that this verifies delegation, not the Haldane gap value itself (3). Items (1) and (5) are informational.

PR #446, opened 2026-05-20
Cards: small-models batch 6 — MG SpinGap + CB + EH delegation + MFI (#381)

sotashimozono:
Adds verify() cards for 4 more hubs:

- **MajumdarGhosh / SpinGap / Infinite**: White-Affleck 1996 PRB 54 9862 DMRG ≈ 0.234J
- **ConformalBootstrap / ConformalWeights / Infinite**: 3D Ising spin op Δσ = 0.5181489 (Kos-Poland-Simmons-Duffin 2014 PRD 86)
- **ExtendedHubbard1D / ChargeGap / Infinite** at V=0: delegation_invariant to Hubbard1D(t, U, μ=U/2)
- **MixedFieldIsing1D / MassGap / Infinite** at default (J=h=1, g=0): TFIM critical line h=J ⇒ Δ=0 (Pfeuty 1970)

Refs #381.

### DIFF
--- a/test/models/quantum/Heisenberg/test_majumdar_ghosh.jl
+++ b/test/models/quantum/Heisenberg/test_majumdar_ghosh.jl
@@ -182,7 +182,10 @@ end
 @testset "MajumdarGhosh — SpinGap White-Affleck (#381 batch 6)" begin
     # Spin-1/2 J1-J2 chain at MG point J2 = J1/2: White-Affleck 1996 DMRG
     # singlet-triplet spin gap Δ_S ≈ 0.234 J.
-    for J in (0.5, 1.0, 2.0)
+    # Single J point: solver returns J * 0.234 from a stored constant;
+    # a multi-J sweep collapses to identical constant-multiplication
+    # residuals on the same code path and adds no discriminating power.
+    let J = 1.0
         verify(
             MajumdarGhosh(; J=J),
             SpinGap(),
--- a/test/models/quantum/Heisenberg/test_majumdar_ghosh.jl
+++ b/test/models/quantum/Heisenberg/test_majumdar_ghosh.jl
@@ -178,3 +178,20 @@ end
         refs=["White-Affleck 1996 DMRG; Eggert 1996: Delta ≈ 0.234 J"],
     )
 end
+# ── additional verification cards (#381 batch 6) ─────────────────────────
+@testset "MajumdarGhosh — SpinGap White-Affleck (#381 batch 6)" begin
+    # Spin-1/2 J1-J2 chain at MG point J2 = J1/2: White-Affleck 1996 DMRG
+    # singlet-triplet spin gap Δ_S ≈ 0.234 J.
+    for J in (0.5, 1.0, 2.0)
+        verify(
+            MajumdarGhosh(; J=J),
+            SpinGap(),
+            Infinite();
+            route=:literature_value,
+            independent=0.234 * J,
+            agree_within=5e-3,
+            refs=["White-Affleck 1996 PRB 54 9862: MG point singlet-triplet spin gap Δ_S ≈ 0.234 J (DMRG)"],
+        )
+    end
+end
+
--- a/test/models/quantum/misc/test_extended_hubbard1d.jl
+++ b/test/models/quantum/misc/test_extended_hubbard1d.jl
@@ -35,17 +35,31 @@ end
 @testset "ExtendedHubbard1D — ChargeGap V=0 delegation to Hubbard1D (#381 batch 6)" begin
     # At V=0 the t-U-V Extended Hubbard chain reduces to the Lieb-Wu
     # Hubbard1D at half-filling. The src delegates ChargeGap directly to
-    # Hubbard1D(t, U, U/2). Cross-check via independent delegate fetch.
+    # Hubbard1D(t, U, U/2). Reference values are precomputed Lieb-Wu
+    # integral values (quadgk rtol≈1e-12) inlined here rather than
+    # re-fetched at test time — a pre-fetch outside verify() would
+    # bypass the verify() failure-handling contract if the upstream
+    # Hubbard1D fetch threw (issue #390 underflow risk at small U).
+    # Provenance: values regenerated via
+    #   QAtlas.fetch(Hubbard1D(; t, U, μ=U/2), ChargeGap(), Infinite())
+    # on Panza @ feat/cards-batch6-v3-381 (2026-05-20).
+    refs_by_pt = Dict(
+        (1.0, 1.0) => 0.005026732898582302,
+        (1.0, 4.0) => 1.2867270220129354,
+        (0.5, 4.0) => 2.339758553732098,
+    )
     for (t, U) in ((1.0, 1.0), (1.0, 4.0), (0.5, 4.0))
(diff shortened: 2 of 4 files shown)


sotashimozono:
## Comprehensive Review — `/review-pr` multi-agent pass

Run on `feat/cards-batch6-v3-381` @ `d1d72b02`. Specialized agents: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`. Findings deduped and ranked by severity, confidence ≥ 80 only.

---

### 🔴 Critical (3)

**C1. `MixedFieldIsing1D` constructor kwargs are wrong — `MethodError` before `verify()` runs.**

```julia
MixedFieldIsing1D(; J=J, h=J, g=0.0)  # ← in PR
```

The struct constructor (verified on Panza, `src/models/quantum/MixedFieldIsing1D/MixedFieldIsing1D.jl`) is:

```julia
function MixedFieldIsing1D(; J::Real=1.0, h_x::Real=1.0, h_z::Real=0.0)
```

No `kwargs...` slurp. Passing `h=` and `g=` throws `MethodError: no method matching MixedFieldIsing1D(...)` at the very first iteration — the test set fails as **errored**, not as a `verify()` mismatch, and the evidence card is never emitted. The intended physics check (TFIM critical line gap = 0) is never executed.

Fix:
```julia
MixedFieldIsing1D(; J=J, h_x=J, h_z=0.0)
```

**C2. `ExtendedHubbard1D` delegation card pre-fetches `ref` outside `verify()`, masking failure modes.**

```julia
ref = QAtlas.fetch(Hubbard1D(; t=t, U=U, μ=U/2), ChargeGap(), Infinite())
verify(ExtendedHubbard1D(...), ChargeGap(), Infinite(); independent=ref, ...)
```

If the `Hubbard1D` fetch throws (dispatch error, integrator divergence, underflow — `#390` showed this is real at small `U`), the exception propagates **before** `verify()` runs. The route's diagnostics, structural classification, and card emission are all bypassed; CI sees an unrelated error, not a tracked verification failure. The chosen sweep `(t, U) ∈ {(1,1),(1,4),(0.5,4)}` avoids `U=0` so this is latent today, but it is a structural anti-pattern that defeats the `:delegation_invariant` route's purpose.

Fix options: (a) move the delegate fetch *inside* `verify()` via a callable, or (b) wrap with `try/catch` that re-raises as a `@test` failure, or (c) replace `independent=ref` with the Lieb-Wu closed-form value at the chosen `(t, U)` and add a separate equality test for the delegation wiring.

**C3. `ConformalBootstrap` card uses wrong `fetch_kw` keyword names; tests silently fall back to default.**

```julia
fetch_kw=(; r=1, s=1)
```

The CB fetch signature (verified, `src/models/classical/ConformalBootstrap/ConformalBootstrap.jl:88`) is:

```julia
fetch(::ConformalBootstrap, ::ConformalWeights, ::Infinite; field::Symbol=:σ, kwargs...)
```

`r` and `s` are MinimalModel/WZW kwargs, not ConformalBootstrap kwargs — they are silently swallowed by `kwargs...`. The card actually exercises the default `field=:σ` branch. By coincidence that is the intended 3D Ising spin-op branch and the value `0.5181489` matches, so the card passes — but for an undocumented reason and via a completely different API surface than the `refs` string implies.

Fix: `fetch_kw=(; field=:σ)` (matching the pre-existing CB verification cards in the same file).

---

### 🟠 Important (4)

**I1. MG SpinGap 3-J sweep is structurally a no-op — three iterations test the same constant multiplication.**

The MG SpinGap solver is `J * _MAJUMDAR_GHOSH_SPIN_GAP_PER_J` (stored constant `0.234`); the card's `independent=0.234 * J` mirrors this exactly. The three iterations `J ∈ {0.5, 1.0, 2.0}` all collapse to the same code path with the same identity assertion — a bug replacing the constant with e.g. `0.233` would be caught by exactly one iteration (the others pass within `5e-3`), not three. The pre-existing `@testset "MajumdarGhosh — SpinGap Δ ≈ 0.234 J"` at `test_majumdar_ghosh.jl:178` already covers this with a tighter `atol=1e-14`.

Action: drop the sweep to a single J or replace with a *different* corroborating route (e.g. cross-check against `J1J2Heisenberg1D` at `J2 = J1/2`).

**I2. MFI MassGap sweep is also structurally a no-op (once C1 is fixed).**

`2|h_x − J| = 2|J − J| = 0.0` exactly for every J, so the three iterations all return `0.0` from the same subtraction. Identical issue to I1.

Action: single J value, or pick a sweep that varies something non-trivial (e.g., `h_x = J + ε` for small ε > 0 to verify the linear scaling above the critical line).

**I3. 3D Ising Δσ value precision mismatch with cited reference.**

`independent=0.5181489` (7 sig figs) matches **Simmons-Duffin 2017** (central value 0.518148806(24)). The PR `refs` string cites only **Kos-Poland-Simmons-Duffin 2014 PRD 86 025022**, whose precision is ~4–5 sig figs (`±0.0001`). At `agree_within=1e-6` the card silently demands 2017 precision while the citation says 2014. A solver implementing the 2014 algorithm would fail the card despite implementing the cited reference correctly.

Fix: add Simmons-Duffin 2017 to `refs` (`"Simmons-Duffin 2017 JHEP 03 086: refined Δσ = 0.518148806(24)"`) OR round `independent` to `0.5182` and relax `agree_within` to `5e-4`.

**I4. EH delegation card is structurally circular — `independent` and `subject` both invoke the same Lieb-Wu integrator.**

`independent=fetch(Hubbard1D(...))` and `subject=fetch(ExtendedHubbard1D(...))` where the EH dispatch internally calls the same Hubbard1D path. The card asserts `1e-12` agreement between two calls to the same function — it tests the EH→H1D delegation wiring, not the physics. This matches the `:delegation_invariant` route's intended `"asserted"` independence class, but the `refs` string should explicitly say so (e.g. `"structural delegation test; physics corroboration is via Hubbard1D verification cards"`) so reviewers don't misread the `1e-12` tolerance as a numerical precision claim.

---

### 🟡 Advisory (3)

**A1. EH `U=0` exclusion needs an explicit comment** (per `#390` precedent — the underflow risk you flagged on Hubbard1D ChargeGap). One-line comment `# U=0 excluded: underflow risk per #390` would prevent the next batch author from extending the sweep into a known broken regime.

**A2. MG `agree_within=5e-3` is absolute, not relative.** At J=0.5 the effective relative tolerance is ~4.3% (independent = 0.117); at J=2.0 it is ~1.1%. Either document the absolute-tolerance intent, or switch to `agree_within = 5e-3 * J` for uniform discriminating power across the sweep.

**A3. MFI MassGap `agree_within=1e-9` is fine after C1 fix** (the closed-form `2|h_x - J|` returns IEEE-exact `0.0` for `h_x == J`), but worth a one-line comment noting the tolerance assumes the analytic-zero return path so future maintainers don't loosen it inadvertently.

---

### Verdict

**Block on merge.** C1, C2, C3 must be fixed before this PR is merge-able:
- C1: the test currently does not execute the intended assertion at all
- C2: silently bypasses verify()'s failure-handling contract
- C3: the assertion passes via the wrong API path

I1, I2, I3, I4 are not crash-class but materially reduce the PR's contribution to verification coverage — recommend fixing in the same revision.

PR #447, opened 2026-05-20
Cards: small-models batch 7 — TB1D Energy + KH F(T→0) + Hubbard Δ_c(U→∞) (#381)

sotashimozono:
Adds 3 verify cards:
- TightBinding1D Energy/Infinite = -2t/π (free fermion, Ashcroft-Mermin 1976)
- KitaevHoneycomb FreeEnergy/Infinite at β→∞ = Energy_GS (delegation_invariant)
- Hubbard1D ChargeGap U→∞ asymptote U - 4t + 8t²ln(2)/U (Lieb-Wu)

Refs #381.


sotashimozono:
## Code Review

### Summary
Adds 3 verify cards: `TightBinding1D` Energy at half-filling (μ=0), `KitaevHoneycomb` FreeEnergy(β→∞) = E_GS, and `Hubbard1D` ChargeGap U→∞ asymptote. Constructor kwargs match the structs (`TightBinding1D(; t, μ=0.0)` defaults — μ=0 half-filling, `KitaevHoneycomb(; Kx, Ky, Kz)`, `Hubbard1D(; t, U, μ)`). Routes `:second_closed_form`, `:delegation_invariant`, `:limiting_case` are all valid members of `_VERIFY_ROUTES`.

### Findings

1. **(Critical — anti-pattern flagged) KH FreeEnergy card pre-fetches `ref_E = QAtlas.fetch(m, Energy(:per_site), Infinite())` outside `verify()`.** This is *exactly* the anti-pattern called out in PR #446 review (C2). If the `fetch(KH, Energy, Infinite)` call throws, mutates global state, or simply returns `NaN` for an exotic `(Kx, Ky, Kz)` combination, the error surfaces *outside* the `verify(...)` boundary — the verify framework never sees it, no JSONL evidence is emitted, and the test fails with a stack trace pointing to the test file rather than a clean `verify` failure card. The whole point of the verification harness is that `independent` should be a *value* known up-front (literature constant, closed-form expression), not a runtime fetch from another QAtlas surface. **Please refactor**: either (a) replace `ref_E` with a hard-coded closed-form (the Kitaev GS energy density has a Bessel-function closed form for the isotropic point, e.g. `Kx=Ky=Kz=1` gives ε₀ ≈ -0.39432); or (b) if the card genuinely tests "free energy delegates to GS energy at T=0," route this through a custom `:delegation_invariant` mechanism inside `verify()` rather than pre-fetching.

2. **(Same anti-pattern: sweep collapse)** Even putting (1) aside, the iteration `(Kx, Ky, Kz) ∈ ((1,1,1), (1,1,2), (0.5,0.5,0.5))` is a *correlated* sweep — for each tuple, both `ref_E` (the LHS) and the fetched FreeEnergy (the RHS) recompute from the same underlying Majorana spectrum. If the src spectrum routine has a bug that affects all (Kx, Ky, Kz), both sides shift in tandem and the verify card still passes. This is structurally similar to #446 I1/I2: parameter sweeps over the *same* code path do not catch correlated errors. The only way to detect such a bug is to compare against a *literature constant* (`ε₀(1,1,1) ≈ -0.39432`) — i.e. fix (1) and the sweep-collapse issue dissolves.

3. **(Hubbard1D constructor & half-filling guard — clean)** The card calls `Hubbard1D(; t=t, U=U, μ=U/2)` — the half-filling particle-hole-symmetric point per the docstring. Confirmed against the struct: `Hubbard1D(t::Float64, U::Float64, μ::Float64)` with kw constructor `Hubbard1D(; t=1.0, U=4.0, μ=2.0)`, half-filling guard satisfied at `μ = U/2`. ✓

4. **(Asymptote choice is correct)** The src Hubbard1D file header explicitly tabulates `Δ_c → U - 4t + 8t² log 2 / U as U → ∞` (Lieb-Wu / Essler 2005 Eq. 6.A.66). The card uses exactly this formula with `log(2)` (natural log, correct — Julia `log` is `ln`). At `U=16, t=1`: asymptote = `16 - 4 + 8·0.6931/16 ≈ 12.347`; exact Lieb-Wu integral at the same U gives ≈ 12.345, residual ~2e-3 → fits comfortably inside `agree_within=2e-2`. ✓ The tolerance choice `2e-2` is well-calibrated for the leading O(1/U²) neglected term.

5. **(Sweep diversity — Hubbard1D good)** The three `(t, U)` tuples are genuinely distinct: different U/t ratios. Unlike (2), the µ-and-asymptote both shift per tuple, so a bug in the leading-order coefficient `8·log(2)` would not cancel across the sweep. Good sweep design, contrast with PR #442's MassGap collapse.

6. **(TightBinding1D — clean)** `TightBinding1D(; t=t)` with default `μ=0.0` puts us at half-filling, `k_F = π/2`, `sin(k_F) = 1`, so `e₀ = -2t/π`. The card uses `independent = -2*t/π` with `agree_within=1e-12` — appropriate machine-precision tolerance for an exact closed form. Sweep `t ∈ (0.5, 1.0, 2.0)` exercises the linear-in-t scaling, which the src `fetch` propagates via the `t::Real=m.t` kwarg path. ✓

7. **(Route classification)** `TB1D` Energy uses `:second_closed_form` — appropriate, the LSM/Ashcroft-Mermin Fourier-integral derivation is independent of the src's piecewise-band-filling implementation. ✓ `Hubbard1D ChargeGap` uses `:limiting_case` — also defensible (U → ∞ limit), and "asserted" status is honest given the Lieb-Wu integral itself is the src truth.

8. **(KH FreeEnergy tolerance)** `agree_within=1e-9` against `BETA=1e6` — at β=10⁶ the thermal correction is ~exp(-β·Δ) with the Kitaev matter-sector gap of order O(K), so the correction is well below machine epsilon. But the `rtol=1e-8` default inside `_kitaev_thermo_infinite` (per `KitaevHoneycomb_thermal.jl`) may dominate the residual, making 1e-9 marginally too tight. Consider `agree_within=1e-7` to leave headroom for the underlying integration rtol. (Independent of the (1) refactor.)

9. **(Refs — minor)** Hubbard1D ref `Essler et al. 2005 Eq. 6.A.66`: the book's Eq. 6.A.66 reports the 1/U expansion of the energy, not the gap directly — the gap asymptote follows from particle-hole-symmetric difference of chemical potentials at half-filling. Consider citing Eq. 6.A.79 or §6.A.4 (large-U gap) for more direct attribution; or simply `Essler et al. 2005, §6.A` to avoid the eq-number risk.

### Verdict
**Request changes.** Item (1) is a real blocker — it's the *same* C2 anti-pattern that #446's review flagged on the EH card, and it should not propagate through batch 7. Fix (1) by replacing `ref_E = QAtlas.fetch(...)` with a hard-coded literature constant (Kitaev isotropic-point ε₀ ≈ -0.39432 from Kitaev 2006 §IV, or one of Knolle/Vidal 2014 numerical values), then (2) dissolves automatically. (8) and (9) are nits. TB1D and Hubbard1D cards are clean and can ship as-is — strongly consider splitting this PR so they don't block on the KH fix.

Issue #445, opened 2026-05-20
XXZ1D thermal kernel: discrepancy vs brute-force ED at finite β


sotashimozono:
## Symptom

The XXZ1D OBC thermal kernel disagrees with full-spectrum ED for `Energy/per_site`, `FreeEnergy`, `ThermalEntropy`, and `SpecificHeat` at every tested `(J, Δ, N, β)`.

The bug-surfacing card is PR #433 (`feat/cards-xxz1d-thermo-OBC-ED-bug-381`): it builds the OBC Hamiltonian from scratch as `H = J · Σᵢ (Sˣᵢ Sˣᵢ₊₁ + Sʸᵢ Sʸᵢ₊₁ + Δ Sᶻᵢ Sᶻᵢ₊₁)` via `chain_hamiltonian(2, N, bond)`, takes the dense spectrum, and computes (E, F, S, C) via the standard `thermo_from_spectrum` log-sum-exp.

## Cross-check (decisive)

At `Δ = 1` the XXZ1D model is exactly Heisenberg1D. Heisenberg1D's OBC thermal kernel matches the same ED reference path to machine precision (PR #432 thermo card passes), so the disagreement at `Δ = 1` in PR #433 cannot come from the ED reference or from the spin-1/2 primitives — it must be inside the XXZ thermal kernel itself.

Suspect surfaces inside `src/models/quantum/XXZ/XXZ_thermal.jl`:

- the bond construction (factor of 1/4 vs 1, or a missing Δ factor on the SzSz term)
- a wrong J prefactor when the kernel reuses cached spectra across different (J, Δ)
- or a mis-application of `_bc_size(bc, kwargs)` that produces an off-by-one chain length

## Refs

- Bug-surfacing card: PR #433
- Upstream pre-existing report: PR #428 (`Cards: XXZ1D Susceptibility{XX,YY,ZZ}/OBC via ED — bug-surfacing (#381)`) — same kernel, same disagreement class
- Trusted ED reference matches at the Heisenberg point: PR #432
- Project tracker: #381 (ED-verify-first sweep)

sotashimozono:
**Verified fixed (2026-06-03).**

Reproduced the bug-surfacing card #433 protocol on current main (`a55043c20`) at the exact reported point (J=1, Δ=1, N=4, β=5):

| Quantity      | Raw ED ref    | QAtlas XXZ1D src   | Diff       |
|---------------|---------------|---------------------|------------|
[...]

Issue #444, opened 2026-05-20
TFIM/PBC SusceptibilityXX disordered phase: parity-sector handling bug


sotashimozono:
## Symptom

The TFIM/PBC thermal kernel disagrees with brute-force ED for `SusceptibilityXX` (and related thermo observables) in the disordered phase (`h > J`). Two independent bug-surfacing cards expose the discrepancy:

- PR #431 — `Cards: TFIM SusceptibilityXX/PBC via ED — bug-surfacing (#381)` (branch `feat/cards-tfim-susXX-PBC-ED-bug-381`)
- PR #436 — `Cards: TFIM Energy + thermo/PBC via ED — bug-surfacing at h>J (#381)` (branch `feat/cards-tfim-thermo-PBC-ED-bug-381`)

Both cards build the TFIM Hamiltonian from scratch, take the full spectrum, and compare the canonical-ensemble observable to the src kernel output. The disagreement is consistent and grows with `h/J`, consistent with a Z2 / parity-sector decomposition error in the PBC thermal kernel (the boundary spin coupling under PBC mixes the +/− parity sectors and a missing sign on the wrap-around bond or a wrong sector projector produces exactly this kind of disordered-phase offset).

## Reference

The brute-force ED card in PR #431 (`test/models/quantum/TFIM/test_TFIM_pbc_thermal.jl` already documents the regime in the original suite) is the trusted reference: it enumerates the full Hilbert space and computes ⟨σˣ σˣ⟩, χ_xx, E, F, S, C from the dense spectrum with no sector tricks. Any discrepancy is therefore on the src side, not the reference side.

## Suggested next step

Audit the PBC parity-sector decomposition in the TFIM thermal kernel — likely either the wrap-around bond sign (Jordan-Wigner string boundary condition) or the per-sector partition-function reweighting is wrong. Heisenberg1D thermal/OBC matches ED exactly, so the bug is PBC- and Z2-parity-specific, not a generic thermo-from-spectrum issue.

## Refs

- Bug-surfacing cards: PR #431, PR #436
- Test file in main suite: `test/models/quantum/TFIM/test_TFIM_pbc_thermal.jl`
- Project tracker: #381 (ED-verify-first sweep)

PR #410, opened 2026-05-20
Cards: S1Heisenberg1D OBC thermal trivial limits (#381)

sotashimozono:
Adds verify() cards for **S1Heisenberg1D/{ThermalEntropy, SpecificHeat, FreeEnergy}/OBC** — 3 uncorroborated hubs.

- T → 0: gapped Haldane GS ⇒ s = 0, c = 0
- T → ∞: spin-1 paramagnet ⇒ s = log 3 per spin, f/N = -log(3)/β, c → 0 (β² tail)

Refs #381.

PR #409, opened 2026-05-20
Cards: XXZ1D OBC thermal trivial limits (#381)

sotashimozono:
Adds verify() cards for **XXZ1D/{ThermalEntropy, SpecificHeat, FreeEnergy}/OBC** — 3 uncorroborated hubs.

- T → 0: s = 0, c = 0
- T → ∞: s = log 2, c → 0 (β² tail), f/N = -log(2)/β

Refs #381.

PR #450, opened 2026-05-21
fix(next CI): resolve s12/s13/s14/s16 — S1H J-swallow bug + TFIM Z2-doublet residue + IsingSquare tolerance

sotashimozono:
## Summary

Resolves the 4 actionable shard failures on `next` (excluding the 3 intentional bug-surfacers s08/s09/s11). Targets the freshly-merged #381 campaign cards.

| Shard | Issue | Severity | Fix |
|---|---|---|---|
| **s13** | S1Heisenberg1D thermo OBC vs ED — solver returned same value for J=0.5 and J=1.0 (J kwarg swallowed) | **REAL BUG** | Pass J via struct constructor instead of fetch_kw |
| **s12** | TFIM SpecificHeat low-T trivial claim — ordered-phase finite-N has Z₂-doublet residue (~0.01) | Card claim wrong | Restrict low-T finite-N OBC to disordered phase (h>J); drop PBC low-T entirely |
| **s14** | TFIM ThermalEntropy low-T trivial claim — ordered-phase finite-N has s ~ log(2)/N (~0.04 at N=16) | Card claim wrong | Same restriction as s12 |
| **s16** | IsingSquare PBC vs brute-force ED — 1e-9 tolerance too tight for transfer-matrix route (~5e-8 diff) | Numerical precision | Relax tolerance 1e-9 → 1e-7 |

### s13 — fetch_kw swallow (same bug pattern as #404, #405, #409, #410)

`S1Heisenberg1D` is a struct with a `J::Float64` field defaulting to 1.0. The card called:

```julia
QAtlas.fetch(S1Heisenberg1D(), Energy(:per_site), OBC(N); J=J, beta=beta)
```

The dispatch:
```julia
function fetch(model::S1Heisenberg1D, ::Energy{:total}, bc::OBC; beta::Real, kwargs...)
    H = _s1_heisenberg_hamiltonian_matrix(model, bc.N)  # reads model.J = 1.0
    ...
end
```

`J` was silently swallowed by `kwargs...`; the actual J of the model was always 1.0. The ED reference (built with the loop J) correctly picked up J ∈ {0.5, 1.0, 2.0}, so the comparison failed at J ≠ 1.

**Fix:** `S1Heisenberg1D(; J=J)` at the constructor, drop J from fetch_kw.

**Local verification (Panza):** All 27 (J, N, β) × 4 observables agree to ≤ 1e-9.

### s12, s14 — Z₂-doublet residue in ordered phase

For TFIM at h<J, the finite-N spectrum has a near-degenerate Z₂-doublet at the bottom: |ψ₊⟩, |ψ₋⟩ split by Δ_doublet ~ h^N (the symmetry-broken cat-state pair). At β·Δ_doublet ~ O(1), the doublet contributes:
- Entropy: s ≈ log(2)/N per site (split into two states still gives entropy)
- Specific heat: c ≈ (β·Δ_doublet)²·exp(-β·Δ_doublet) — NOT exponentially small

Reproduced locally:

```
SpecificHeat (β=100):
  OBC N=8  (J=1, h=0.5) c = 0.00986    (FAIL — card expected 0)
  PBC N=8  (J=1, h=0.5) c = 0.000659   (FAIL)
  OBC N=8  (J=1, h=2)   c = 4.5e-92    (pass — disordered)
  Infinite (J=1, h=0.5) c = 1.1e-41    (pass — Pfeuty exact 0)

ThermalEntropy (β=1000):
  OBC N=16 (J=1, h=0.5) s = 0.0433     (FAIL — card expected 0; matches log(2)/16 ≈ 0.0433)
  OBC N=8  (J=1, h=2)   s = 0.0        (pass — disordered)
```

**Fix:** restrict the s=0 / c=0 low-T claims on finite-N OBC to h>J (disordered: field-induced gap → exponentially suppressed). PBC low-T cards dropped entirely because neither phase qualifies (ordered → doublet residue, disordered → known parity-sector bug). Infinite() low-T cards untouched (Pfeuty gives exact 0 in both phases).

Two `verify()` cards drop from INVENTORY (386 → 384): the TFIM PBC low-T SpecificHeat and ThermalEntropy s=0 claims that had no valid regime.

### s16 — tolerance bump

Brute-force ED on (Lx, Ly) = (3, 3) PBC vs IsingSquare transfer-matrix solver: agreement to ~5e-8 across all 3 observables. The 1e-9 tolerance was over-tight — relaxed to 1e-7. Order-of-magnitude bugs still detectable by O(1).

## Out of scope

- **s08, s09, s11**: intentional bug-surfacing cards tracked in issues #444, #445, #423. These will fail CI until the upstream solver bugs are fixed; that is the design intent.

## Test plan

- [x] s13 local verification: 27 × 4 = 108 comparisons all agree to < 1e-9 with the fix.
- [x] s12 / s14 local reproduction confirms the residue is intrinsic to ordered-phase finite-N and not a solver bug.
- [x] JuliaFormatter `format(".", overwrite=false)` returns `true`.
- [x] INVENTORY drift guard regenerated.
- [ ] CI on `next` after merge: 4 shards (s12, s13, s14, s16) flip to green; 3 bug-surfacers (s08, s09, s11) remain red as intended.


sotashimozono:
## Comprehensive Review — `/review-pr` multi-agent pass

Run on `fix/shard-failures-next` @ `f8880f82` (6 commits, 30 files, +144/-177). Specialized agents: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`. Findings deduped, severity-ranked, confidence ≥ 80 only.

---

### 🔴 Critical (2)

**C1. Hubbard1D `t² → t` fix has zero regression coverage at `t ≠ 1`** (flagged by all 3 agents)

The core solver fix in `src/models/quantum/Hubbard1D/Hubbard1D.jl`:

```julia
return -4.0 * t^2 * integral   # was — dimensionally wrong, only coincided at t=1
return -4.0 * t   * integral   # now — correct Lieb-Wu 1968
```

is locked by exactly ONE multi-`t` regression card: `test_hubbard1d_gsed_U0_batch.jl`, which sweeps `t ∈ (0.5, 1.0, 2.0)` against the **U → 0** free-fermion textbook value. Every other existing GSED test in `test_hubbard1d.jl` is pinned at `t = 1` where the bug is invisible (t² = t).

The **finite-U regime** — where the formula actually has nontrivial `t`-dependence — has zero `t ≠ 1` coverage. A future regression that re-introduces the `t²` factor at finite U would pass all current tests except the U → 0 batch, and a partial revert (e.g. only the strong-coupling branch) would slip silently.

**Action:** Add a `verify()` card for `GroundStateEnergyDensity` at the **U → ∞** asymptote `E₀/N → -4 t² log(2) / U` (the one limit where `t` vs `t²` is unambiguous at finite coupling) over `t ∈ {0.5, 2.0}` with `U = 100 t`. This locks both ends of the formula's `t`-dependence.

**C2. Same J/Δ-swallow pattern still latent in `test_heisenberg1d_obc_thermo_ED_batch.jl`** + 16 src dispatch sites

This PR fixes the swallow bug in s13 (S1Heisenberg) and s8 (XXZ) — but `test/models/quantum/Heisenberg/test_heisenberg1d_obc_thermo_ED_batch.jl` (lines 31, 44, 55, 66) has 4 identical instances of the exact pattern:

```julia
verify(
    Heisenberg1D(),                 # J not passed via struct
    Energy(:per_site),
    OBC(N);
    fetch_kw=(; J=J, beta=beta),    # J swallowed by dispatch kwargs...
    ...
)
```

CI may be currently green because the loop happens to include `J=1.0` (which matches the default), masking the bug at `J ≠ 1`.

**Plus** — the `silent-failure-hunter` audit identified **16 src dispatch sites** with the same `J::Real=m.J, ..., kwargs...` shape (so any `fetch_kw=(; J=...)` from a card silently swallows). All cards that call these models via `fetch_kw` are candidates for the same false-corroboration class:

```
S1XXZ1D, S1AnisotropicD1D, KitaevHeisenberg, LongRangeIsing1D,
MajumdarGhosh, LongRangeXY1D, MixedFieldIsing1D, DMIHeisenberg1D,
KagomeHeisenbergAFM, AKLT2D, Cluster1D, ShastrySutherland
```

**Action:** apply the same struct-constructor fix to `test_heisenberg1d_obc_thermo_ED_batch.jl` in this PR (it shares the same class), and file a follow-up audit task for the 16 src sites + any cards that pass `J`/`Δ`/`D`/etc. via `fetch_kw`.

---

### 🟠 Important (3)

**I1. Dropped S1H low-T cards leave the spin-1 OBC low-T regime entirely uncovered**

The PR correctly removes two `independent = 0.0` cards from `test_s1heisenberg1d_obc_thermal_batch.jl` (finite-N OBC has edge-multiplet residue `s ≈ log(3)/N`, not 0). But it does NOT replace them with the correct finite-N form. The S1H OBC low-T regime now has no card. A future solver regression in that regime would not be caught.

**Action:** add the correct cards:
```julia
verify(
    S1Heisenberg1D(; J=J), ThermalEntropy(), OBC(N);
    independent = log(3)/N,                    # edge-triplet residue
    agree_within = 1e-9,                        # exact in IEEE 754
    route = :limiting_case,
    refs = ["S1H OBC finite-N: edge S=1 triplet GS ⇒ s_per_site = log(3)/N"],
    fetch_kw = (; beta = LOW_T_BETA),
)
```

(and the analogous `c = (β·Δ_edge)²·exp(-β·Δ_edge)` for SpecificHeat, or a numerical pin from the ED batch).

**I2. `verify()` framework's new `expected_fail` branch has no unit test**

`test/util_verify/` covers the regular `@test` path; the new `@test_broken` branch added in this PR is exercised by 3 cards in production but never tested directly. A future change to the framework (e.g. renaming the kwarg or breaking the dispatch on `expected_fail`) would silently no-op the broken status.

**Action:** add `test/util_verify/test_verify_expected_fail.jl` covering: (a) `expected_fail=true` + actually-failing assertion → test passes (broken); (b) `expected_fail=true` + actually-passing assertion → test errors with "Unexpected Pass"; (c) JSONL emit (`QATLAS_EMIT=1`) records the flag (see I3).

**I3. `expected_fail` flag not recorded in emitted JSONL card**

The emit block in `verify.jl` computes `status` independently via `isapprox(best, subj; atol=agree_within)`. With `expected_fail=true`, a failing card emits `status="fail"` with NO indication that the failure was expected, and a passing card emits `status="pass"` with no flag for promotion.

**Action:** add `"expected_fail":true` to the emitted card when set, so dashboards can distinguish "tracked broken card" from "unintentional regression" and surface candidates for promotion back to `@test`.

---

### 🟡 Advisory (3)

**A1. `@test_broken` loop-granularity partial-fix masking** (s16 TFIM PBC conditional)

The `expected_fail=(h > J)` pattern works for the binary ordered/disordered split, but if a future kernel fix for #444 only fixes SOME of the `(h > J)` × N × β iterations, the fixed ones emit "Unexpected Pass" warnings while the unfixed ones silently stay broken. There is no mechanism to require ALL `expected_fail` assertions to flip together.

**Mitigation already in place:** conditional `expected_fail=(h > J)` splits the regime cleanly. Worth tracking explicitly in #444 that the demarcation needs to remain coherent.

**A2. INVENTORY hub-key mismatch: `ShastrySutherland` records two prefixes**

The `QAtlas.ShastrySutherland` prefix fix in `test_shastry_sutherland.jl` causes INVENTORY rows to differ in key (`QAtlas.ShastrySutherland/...` vs `ShastrySutherland/...`). If the docs-hub generator strips module prefixes (or vice-versa), the same hub appears twice in INVENTORY but once in the docs page, or zero times after normalization.

**Action:** verify the INVENTORY scanner normalizes `QAtlas.X` → `X` for hub keys. If not, normalize at scan time.

**A3. `_verify_commit()` and `scan_file` bare-catch swallow diagnostics**

- `verify.jl:64` — `try ... catch ... return "unknown" end` discards exception. Cards from broken CI environments are indistinguishable from genuine emissions. Low impact (evidence trail only).
- `AtlasInventory.jl:451-469` — `PARSE_FAILS` records only `typeof(err)` not `string(err)`. Lost line numbers / parse details. The `test_majumdar_ghosh.jl` parse-error episode earlier in this campaign would have been much faster to diagnose with full error messages.

**Action:** change to `catch err; @debug "..." exception=err; ... end` patterns; record `string(err)` in `PARSE_FAILS`.

---

### Positive observations

- **Hubbard1D fix is dimensionally correct** and well-explained in the commit message (`-4 t × integral` reproduces both U → 0 textbook -4t/π AND U → ∞ textbook -4t² log(2)/U).
- **s8 was a real misdiagnosis** — what looked like an XXZ thermal kernel SU(2) violation (#445) was actually the same J/Δ swallow. The probe (ED vs `fetch_kw` vs `struct constructor`) cleanly establishes this; #445 can close.
- **`expected_fail` mechanism design is sound** — `@test_broken` promotes via "Unexpected Pass" → the bug-surface signal survives end-to-end when the underlying bug is fixed.
- **Conditional `expected_fail=(h > J)`** for the TFIM PBC thermo is the right granularity for the regime split.
- **108-point S1H ED batch (27 × 4)** with `agree_within=1e-9` after the constructor fix is exemplary structural coverage.
- **The 3 dropped cards** (S1H 2 + INVENTORY-implicit) are honest about the physics (no spurious "trivial = 0" claims at finite-N OBC) — the right call even though it creates coverage debt.

---

### Verdict

**Merge-able with C1/C2 follow-up.**

- **C1 (Hubbard1D regression coverage at t≠1)** and **C2 (Heisenberg1D ED batch with same J-swallow)** are the only critical-class findings. **C2 in particular should be added to THIS PR** since it's exactly the same class as the bugs being fixed; landing the same fix at the same time keeps the audit closed.
- **C1 (the U→∞ regression card)** is small (one verify call) and can be added to this PR or a small follow-up.
- I1/I2/I3 + advisory items can be follow-up PRs without blocking the merge.

sotashimozono:
## Comprehensive Review (Round 2) — `/review-pr` on new commits

Run on `fix/shard-failures-next` @ `f1aa5e8e` (3 new commits since prior review: `836a0571` C1+I1, `0f2ee3eb` I2+I3, `f1aa5e8e` format). Specialized agents: `code-reviewer`, `pr-test-analyzer`, `silent-failure-hunter`, `comment-analyzer`. Findings deduped, severity-ranked, confidence ≥ 80 only. Findings from the prior round (C2 src audit, A3 bare-catch) intentionally omitted.

---

### 🟠 Important (3)

**I1. S1H replacement card: misleading "measured locally" attribution + redundant N=3/4/5 lookup table**

`test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_thermal_batch.jl` lines 25–37

The new comment block says values were "measured locally (Panza, J ∈ {0.5, 1.0, 2.0}, N ∈ {3, 4, 5})" and prints `N=3: 0.3662 / N=4: 0 / N=5: 0.2197`. But these are not empirical measurements — they are floating-point evaluations of the parity formula `iseven(N) ? 0.0 : log(3)/N`, which the verify card itself encodes. Two issues:

- Attribution overstates the empirical content: a reader will look for the script that "measured" these values on Panza and not find one. The formula is the analytic claim; the values are evaluations of it.
- The lookup table adds zero information beyond `log(3)/3` / `log(3)/5` and carries rot risk: if `LOW_T_BETA` or `N` ranges change, the printed numbers will silently drift from the test parameters.

**Action:** Drop the three `N=…` lookup lines. Rewrite the attribution as "edge-multiplet GS degeneracy: `2S_edge + 1 = 3` for two unpaired boundary S=1/2 spins on odd N, paired into a singlet for even N; entropy per site `s = log(GS deg)/N`." That is the actual physics; cite Kennedy 1990 (Commun. Math. Phys.) for the boundary-S=1/2 result.

**I2. Hubbard1D U→∞ card: comment misstates the sub-leading correction order**

`test/models/quantum/misc/test_hubbard1d_gsed_U0_batch.jl` lines 44–45

The comment reads: "agree_within = 1e-4 covers the leading sub-leading correction (relative ~ t/U ~ 1% at U/t = 100, abs ~ 5e-5 at the worst sweep point)". Two problems:

- "leading sub-leading" is self-contradictory phrasing.
- The actual next-to-leading correction in the strong-coupling Lieb-Wu expansion is `O(t⁴/U³)`, not `O(t/U)`. At U=100t the relative correction is `~(t/U)²·|e₀| ≈ 1.4×10⁻⁶` (t=0.5) to `5.6×10⁻⁶` (t=2.0), so `agree_within=1e-4` is even safer than the comment claims — but the comment overstates the residual by ~100×.

The tolerance is fine; the rationale comment is wrong and will mislead the next person who tries to tighten or relax it. Verified independently by both `code-reviewer` and `pr-test-analyzer` with concrete numerical traces.

**Action:** Rewrite to "next-to-leading correction is `O(t⁴/U³) ≈ (t/U)²·|leading| ~ 5×10⁻⁶` at U=100t, comfortably inside `1e-4`."

**I3. `_CaptureSet` missing `Test.results(...)` override → silent under-count in CI summary**

`test/util_verify/test_verify_expected_fail.jl` lines 35–41

`Test.AbstractTestSet` provides a default `results(::AbstractTestSet) = ()`. `_CaptureSet` stores results in `.results` field but does not override `Test.results(ts::_CaptureSet)`. The file's own assertions access `res.results` directly so they are correct, but `Test.print_test_results` / `Test.get_test_counts` (run by the verbose CI summary) call `results(ts)` and see `()`. Net effect: Broken / Error / Pass records inside `_CaptureSet` blocks do not appear in the suite-level CI count.

Not a test correctness failure today, but if a future CI gate counts broken tests to decide merge eligibility (or surfaces them in a diff vs prior run), the `_CaptureSet` subtrees are invisible.

**Action:** One line — `Test.results(ts::_CaptureSet) = ts.results` next to the existing `Test.record` / `Test.finish` overrides.

---

### 🟡 Advisory (4)

**A1. Hubbard1D U→∞ card route should be `:literature_value`, not `:limiting_case`**

`route=:limiting_case` maps to `independence="asserted"` in the emitted JSONL. The refs cite Lieb-Wu 1968 + Essler et al. 2005 — published Bethe-ansatz analytic results. The correct route for a published numeric cross-check is `:literature_value`, which would emit `independence="structural"` and surface this card in evidence-trail audits that filter for structural corroboration.

**Action:** Change `route=:limiting_case` → `route=:literature_value` for the U→∞ verify card. Same change worth considering for the existing U→0 sister card on the same hub.

**A2. S1H T→0 SpecificHeat counterpart not restored — asymmetric with restored entropy card**

`test/models/quantum/Heisenberg/test_s1heisenberg1d_obc_thermal_batch.jl`

The PR commit that originally removed the two trivial-limit cards (`f8880f82`) dropped BOTH `ThermalEntropy` and `SpecificHeat`. This PR restores only the entropy with the parity-correct `iseven(N) ? 0 : log(3)/N` form. The matching `SpecificHeat` is `c = 0` for all N (gapped finite system, both singlet and triplet GS), and the existing S=1/2 analogue `test_heisenberg1d_obc_thermal_batch.jl` has both T→0 entropy and T→0 specific-heat cards. The asymmetry is a documentation hazard: a future reader will wonder whether the absent card is omitted because the physics is non-trivial or merely because it was forgotten.

**Action:** Add the symmetric `SpecificHeat` card with `independent=0.0, agree_within=1e-9, route=:limiting_case` and a one-line ref note ("gapped GS for both parities ⇒ c → 0 at T → 0").

**A3. `test_verify_expected_fail.jl`: 3 small coverage gaps worth one cycle**

The 14 assertions cover the three header-stated contracts and the default `expected_fail=false` happy path. Three additional one-line guards would noticeably increase the file's regression value:

- `expected_fail=false` + mismatch → `Test.Fail` (the normal regression-fail path is the production hot path; not tested).
- A single `JSON3.read(line)` or stdlib `JSON.parse(line)` on the emitted JSONL to confirm structural validity (the current `occursin` substring checks pass on raw-string concatenations and miss a malformed top-level token).
- Contract (b) lacks the symmetric `count(r -> r isa Test.Broken, res.results) == 0` zero-assertion that contract (a) has.

**A4. `expected_fail` boolean serialised by inline ternary, not via a `_json_bool` helper**

`test/util/verify.jl` lines 239–240. The output `true`/`false` is valid JSON, but every other scalar field goes through `_json_str` / `_json_num`. A future contributor adding the next boolean field has even odds of `_json_str(b)` (which produces a quoted string, breaking typed parsers). One-line fix: add `_json_bool(b::Bool) = b ? "true" : "false"` next to `_json_num` and route the field through it.

---

### Positive observations

- **The U→∞ asymptote card definitively catches the original `t²→t` bug.** Numerical traces: at t=0.5 / U=50, |fixed src − asymptote| ≈ 5×10⁻⁶; |bug src − asymptote| ≈ 7×10⁻³ — bug exceeds tolerance by ~700×. The structural argument that "t² is in the asymptote formula anyway so the bug might survive" is wrong: the inner integral carries one extra factor of t/U, so the buggy outer-t² prefactor produces t³/U scaling, not t²/U.
- **The S1H parity formula is physically correct** for the OBC S=1 chain edge-state spectrum (Kennedy 1990 boundary-S=1/2 multiplets) and the t=0 thermal kernel exactly reproduces it (β·Δ_Haldane ≥ 2×10⁵ at LOW_T_BETA=1e6, all excited-state weights underflow to 0).
- **`expected_fail` JSON emit is correctly placed and serialised** as a JSON boolean token; existing `test_verify_json.jl` regression guard is unaffected (field-order agnostic). The new `test_verify_expected_fail.jl` test file passes 14/14 assertions locally.
- **JuliaFormatter clean** on the new test file. CI format-check should be green.

---

### Verdict

**Merge-able with I1+I2 comment fixes recommended.**

- I1 + I2 are pure comment / documentation fixes — single Edit per file, two minutes of work. Worth landing now because both will mislead the next reader.
- I3 (`Test.results` override) is a one-line addition; land alongside I1+I2.
- A1–A4 are honest enhancements with no failure surface today; can land in a follow-up or be folded in if cheap.
- All "critical" candidates from the agent passes either dedupe back to the prior review's tracked items (C2 src-side audit) or downgraded after verification (the "structurally circular" claim against the S1H card is wrong — the formula is analytic, not measurement-derived; the comment wording is what gave that impression, addressed by I1).

Issue #702, opened 2026-06-11
@measured + phase-diagram registry: experimental anchors and phase regions (future)


sotashimozono:
Part of #697 (meta-registry roadmap). Deliberately FUTURE — parked here so the design space is recorded.

## @measured — experimental anchors

Attach experimental values with uncertainties to hubs:

```julia
@measured Heisenberg1D ZZStructureFactor Infinite value=... err=... at=(; q=π, ω=...) refs=["NeutronExpt..."]
```

- Different epistemic status from exact/ED rows: comparison is `|theory − exp| ≲ kσ`, never exact.
- Gives the atlas its theory↔experiment face (the "new kind of textbook" axis), and a reliability tier between cited-only and verified.

## Phase-diagram registry

Generalize the point predicates (`at = (m -> isapprox(m.h, m.J))`) of `@realizes` into phase *regions* with labelled transitions:

```julia
@phase TFIM :ordered    region = (m -> m.h < m.J)
@phase TFIM :paramagnet region = (m -> m.h > m.J)
@transition TFIM (:ordered, :paramagnet) realizes=:Ising order=:continuous
```

Coherence: every transition between registered phases must either carry a universality class or be declared first-order; order parameters live in phases; `realizes` edges live on transitions. This subsumes the current critical-locus predicates and makes "phase diagram pages" in the atlas generatable.

Both depend on the recipe + experience from @identity/@dual (#697 children); do not start before those land.

Issue #701, opened 2026-06-11
@limits_to: asymptotic limit edges with convergence-sequence test generation


sotashimozono:
Part of #697 (meta-registry roadmap).

## What

Asymptotic limits between models, distinct from `@reduces` (exact delegation at a point in parameter space):

```julia
@limits_to XXZ1D Heisenberg1D as = :(Δ → 1⁺)   # isotropic limit
@limits_to XXZ1D IsingChain   as = :(Δ → ∞)    # Ising anisotropy limit
@limits_to TFIM  IsingSquare  as = :(via path-integral / β→∞ mapping)
```

with optional convergence-rate metadata.

## Why

PR #691 hand-wrote exactly this pattern: the XXZ exact 2-spinon DSF is checked to converge to the (independently exact) Heisenberg formula along Δ ∈ (1.001, 1.0001, 1.00001). That is a *generated-test shape*: edge + quantity implemented on both sides ⟹ emit a convergence-sequence test (sequence of parameter values, tolerance schedule from the rate metadata).

Asymptotic edges also extend selective CI's blast radius and document, in the graph, which models are IR/strong-coupling shadows of which.

## Notes

- Needs care with kwargs alignment between the two sides (q, ω, β conventions) — reuse the `quantity_map` mechanism from the `@dual` issue.
- Generator emits the sequence test only when BOTH endpoints have non-delegating implementations for the quantity (otherwise it is circular).

Implementation recipe as in #697.

sotashimozono:
## Measured state: the machinery is shipped and has **zero users**

`src/core/limits.jl` defines `LIMIT_EDGES`, the `:limit` generator is registered, and everything is wired. Then:

```
generated checks by kind:
  identity  94
  bound    146
  response 306
  region    55
  dual      52
  symmetry   5
  limits     0     ← here
```

`LIMIT_EDGES` is empty. So this issue is **not** "build the mechanism" any more — it is "declare the first edge", which is a different and much smaller job.

## Why that distinction matters enough to write down

A registered generator that produces zero checks is **invisible**. It does not fail, it does not warn, and it adds a name to the kind list that reads like coverage. The bound generator had the same shape until #822 gave it edges; the difference is that someone went looking. Nothing in CI would have told either of us.

[...]

Issue #700, opened 2026-06-11
@symmetry: model symmetry attributes + LSM-type coherence checks


sotashimozono:
Part of #697 (meta-registry roadmap). Related: #689 (symmetry axis of the 1D roadmap).

## What

Node *attributes* (not edges): register each model's symmetry data declaratively —

```julia
@symmetry Heisenberg1D internal=:SU2 translation=true time_reversal=true filling=1//2
@symmetry XXZ1D       internal=:U1  translation=true time_reversal=true filling=1//2
@symmetry TFIM        internal=:Z2  translation=true
```

## Why — the graph learns physics no-gos

This is the watershed between *catalog* and *inference engine*. With symmetry + filling registered, Lieb-Schultz-Mattis-type theorems become **static coherence checks over the registry itself** (no test execution):

- LSM: `internal ⊇ SU(2)` (or U(1) at fractional filling) + `translation` ⟹ a registered `MassGap > 0` at `Infinite` with a unique symmetric ground state is a coherence `:error`.
- Kramers: `time_reversal` + half-integer spin ⟹ degeneracy constraints vs registered spectra.
- Selection rules: symmetry ⟹ which structure-factor channels may be nonzero.

A violation means either a registry mistake or a genuinely interesting claim — both worth surfacing. This extends the C1–C9 coherence family (cf. C9 orphan classes, #661) with the first checks that encode *theorems* rather than bookkeeping invariants.

## Scope discipline

Do NOT backfill all models at once. Land the store + ONE coherence check (LSM for SU(2)/U(1) chains) + the ~6 models needed to exercise it, then grow attribute coverage as models are touched.

Implementation recipe as in #697 (store + macro + query + static checks; no test generator needed initially — checks run in the L0 coherence gate).

sotashimozono:
Shipped. The `:symmetry` generator is registered and produces **5 checks** today; model symmetry attributes are declared and consumed by the identity family mode (`requires_internal = :SU2`).

The parameter-conditional half — `at` predicates so `XXZ1D` at `Δ = 1` is recognised as SU(2) — is tracked separately in #706 and is not lost by closing this. Closing.

Issue #699, opened 2026-06-11
@dual / @coincides: duality edges with parameter maps + automatic cross-implementation checks


sotashimozono:
Part of #697 (meta-registry roadmap).

## What

Declare bijective, parameter-mapped equivalences between models (or between a model and a universality-level description), distinct from `@reduces` (exact delegation at a point):

```julia
@dual TFIM TFIM        param_map = (m -> TFIM(; J=m.h, h=m.J))          kind=:kramers_wannier
@dual TFIM Kitaev1D    param_map = (m -> Kitaev1D(; t=..., μ=...))      kind=:jordan_wigner
@dual XXZ1D CompactBoson param_map = (m -> (; K=luttinger_K(m.Δ)))      kind=:bosonization
@coincides Universality{:Heisenberg} Universality{:XY} at = (; K=1.0)   # SU(2)₁ = self-dual boson
```

## Why — concrete incident

The SU(2)₁ Heisenberg boundary entropy was registered with wrong values (free→0, fixed→+¼log2; correct: both Cardy states g = 2^{-1/4}, log g = −¼log2). The test parroted the source's claims, so CI was green — **circular verification**. The bug was caught only by the hand-derived consistency requirement "XY at the self-dual radius K=1 must equal the SU(2)₁ Cardy states", which is now a hand-written cross-check in the test. A declared `@coincides` edge would have *generated* that check for every quantity both sides implement.

## Test generation

For each dual edge and each quantity implemented on both sides: fetch on side A, map parameters, fetch on side B, compare (with the edge's quantity_map handling operator renormalisation, e.g. spin vs Pauli conventions, velocity factors). These are *independent-implementation* cross-checks — the strongest verification class available without external data, and they directly justify `reliability=:high`.

## Coherence

- dangling endpoints; non-invertible param_map at registered example points;
- a `@coincides` edge between universality classes requires equal central charge (cheap static check).

Implementation recipe as in #697 (store + macro + query + generator).

sotashimozono:
Shipped. `src/core/duality.jl` carries the `DUALITIES` store with parameter maps, and the `:dual` generator produces **52 checks** today. Closing.

PR #704, opened 2026-06-11
feat(registry): constraint-edge kernel — @symmetry / @identity / @dual / @limits_to with generated checks (#697)

### LINKED ISSUE #696: test: validate @realizes reference citations in test_references_bib.jl
## Problem

\`test/core/test_references_bib.jl\` validates that every bibkey cited via \`@register\` exists in \`docs/references.bib\`, but it does **not** validate bibkeys cited via \`@realizes\`. This gap allowed several incorrect/missing citations to slip through CI undetected — discovered during the PR #691–#695 review cycle:

| PR | Broken citation found |
|----|-----------------------|
| #695 | \`Cardy1984\` used for Lee-Yang edge (correct key: \`Cardy1985\`, PRL 54, 1354) |
| #695 | \`Fisher1992\`, \`Fisher1995\`, \`RefaelMoore2004\` missing from \`docs/references.bib\` |
| #694 | \`Cardy1986\` missing from \`docs/references.bib\` |

None of these were caught by CI because \`QAtlas.REALIZES\` is never consulted by the reference test.

## Root Cause

\`test_references_bib.jl\` iterates only \`QAtlas.REGISTRY\` (from \`@register\`). The \`QAtlas.REALIZES\` registry (from \`@realizes\`) is never checked, leaving all its \`references = [...]\` arrays unvalidated.

## Proposed Fix

Extend the loop to also iterate \`QAtlas.REALIZES\`:

\`\`\`julia
for (key, entry) in QAtlas.REALIZES
    for ref in get(entry, :references, String[])
        @test ref ∈ bib_keys  "REALIZES entry $key: bibkey \"$ref\" not found in references.bib"
    end
end
\`\`\`

## Acceptance Criteria

- [ ] \`test_references_bib.jl\` iterates both \`QAtlas.REGISTRY\` and \`QAtlas.REALIZES\`
- [ ] All bibkeys in both registries resolve in \`docs/references.bib\`
- [ ] CI catches future broken \`@realizes\` citations at the PR level

sotashimozono:
## Summary

Implements the **constraint-edge layer** of the meta-registry roadmap (#697) as ONE shared kernel + four thin edge types, instead of five disjoint copies of the store/macro/query/coherence/generator recipe. Also lands the first slice of the quantity taxonomy (#690) and closes the `@realizes` reference-validation gap (#696) as a side effect of the kernel design.

### The unifying design decision

#697 prescribed per edge type: *store + macro + query API + static coherence + test generator*. This PR implements those shared mechanics **once**, in `src/core/constraints.jl`:

- **`EDGE_STORES` self-registration** — every declarative store (the legacy `REGISTRY`/`REALIZES`/`REDUCES`/`ABOUT` and each new constraint store) registers itself with `references`/location accessors. C1 reference integrity is now ONE loop over the registration: a future store (e.g. `@measured`, #702) is covered by C1 and `test_references_bib.jl` the moment it registers — no more editing `coherence.jl` per store (this is what closes #696).
- **`GeneratedCheck` protocol** — the common output type of every constraint generator. `generated_checks()` enumerates (edge × implementations present in `REGISTRY`) deterministically; `test/generated/` runs the result. A new edge type buys mechanical test generation by registering one generator function.
- **Shared hub enumeration** — "which (model, bc) implement ALL of these quantities?", including the **independence filter** (non-delegating rows only) that structurally kills the circular verification behind the SU(2)₁ boundary-entropy incident, and the Energy-granularity equivalence (per-site/total routing is automatic by design).

### Edge types (each ~1 file over the kernel)

| Edge | Issue | Store + macro | Static coherence | Generated checks |
|---|---|---|---|---|
| `@symmetry` | #700 | model symmetry profiles (internal group, translation, T-reversal, site spin, declared gap/degeneracy) | **C10: LSM as a theorem-check** — gapped+unique on a translation-invariant half-odd-integer chain with SU(2)/U(1) ⇒ `:error` | (gates identity generation) |
| `@identity` | #698 | quantity↔quantity relations; tuple mode + **family-isotropy mode over the #690 taxonomy** | C11: unexercised edges self-report as `:gap` | Gibbs `ε = f + T·s` on **24 hubs × 3 β** with zero hand-written tests; SU(2) isotropy (χ, m) gated on `@symmetry` |
| `@dual` | #699 | parameter-mapped equivalences with explicit quantity allowlists + per-quantity `value_map` | C12: param_map type/involution sanity on examples; missing/delegating endpoint rows ⇒ `:gap` | TFIM Kramers–Wannier self-duality + TFIM↔Kitaev1D Jordan–Wigner: 52 independent-implementation cross-checks |
| `@limits_to` | #701 | asymptotic limits (distinct from `@reduces`) with monotone approach sequences + rate metadata | C13: param exists, endpoints present + independent | XXZ → Heisenberg Δ→1⁻ convergence sequence (the generated form of #691's hand-written pattern) |

### Quantity taxonomy (#690, first slice)

8 abstract family supertypes (`AbstractSusceptibility`, `AbstractMagnetization`, `AbstractThermalPotential`, `AbstractGap`, …) inserted between the concrete leaves and `AbstractQuantity` — purely additive, all dispatch unchanged — plus the `component` trait (`:xx`/`:x`/`:mass`/…). This is what lets the SU(2) isotropy identity be declared **once** against `AbstractSusceptibility` instead of per-leaf, and replaces the harness's hand-coded `is_su2_symmetric` model filter with a registry query (`@symmetry` → `models_with_symmetry(:SU2)`).

### Physics verified before declaring

- **Jordan–Wigner TFIM ↔ Kitaev1D**: verified numerically that the implementations agree to machine precision on every listed quantity (the wire's particle-hole-symmetric convention absorbs the naive `−h·N` offset — the per-quantity `value_map` slot exists for maps where such offsets survive).
- **Kramers–Wannier**: `ε(k) = 2√(J² + h² − 2Jh cos k)` is J↔h-symmetric, so the thermal scalars + gap are on the allowlist; the order parameter (maps to the disorder operator) deliberately is not.
- **XXZ → Heisenberg**: approach from the planar side (Δ→1⁻) because the gapped-regime e₀ branch is not implemented (returns NaN); Yang–Yang integral (source) vs closed Hulthén value (target) are genuinely independent implementations.

### Knowledge-graph / atlas improvements

`docs/atlas/graph_export.jl` now draws the meta relations — `reduces`, `dual`, `limits_to` (model↔model) and `identity` (quantity↔quantity) — so the Gibbs F–E–S triangle and the χ component family render as connected structure instead of disconnected leaves, and the model network shows delegation routes, dualities and asymptotic shadows.

### Test/CI integration

- `test/generated/` (new universe dir): one runner per kind; declared exclusions surface as visible `:skip`s, never silent drops.
- `test/core/test_constraints.jl`: kernel/store/macro/query unit tests incl. synthetic LSM violations, broken duality maps, determinism of `generated_checks()`.
- C10–C13 wired into `coherence_report()`; `errors = 0` stays the hard invariant (verified locally: 0 findings).
- `test/core/test_references_bib.jl` now scans **every** edge store via `EDGE_STORES` (#696).

### Deliberately not in this PR

- `@measured` + phase-diagram registry (#702 — explicitly parked; slots into the same kernel).
- `@coincides` between universality classes (the SU(2)₁ = self-dual-boson check needs kwargs-mapped universality fetches; the store design accommodates it).
- Derivative-form identities (`c_v = -β²∂ε/∂β`) — need a derivative engine src/ doesn't depend on; they stay in the hand-written harness, which remains the reference per #698's migration plan.
- INVENTORY-style static drift guard for the new stores (rows can only enter via the macros; revisit with the selective-CI work).

Refs #697, #698, #699, #700, #701, #690. Closes #696.


sotashimozono:
## Review pass — all findings fixed in c0a5166626

Focus areas: design soundness / extension cost / runtime weight / test splitting / duplication.

| # | Finding | Fix |
|---|---|---|
| 1 | `test_identity_checks.jl` = **3108 s** (86 % of shard s06's 1h08m): `finite_N=8` put spin-1 hubs at 3⁸ = 6561-dim dense ED **per fetch**, no caching | edges declare `finite_N=6` (hand-written files use OBC(4)/OBC(6); the identities are exact at any N) → suite drops to seconds. File split into **gibbs / isotropy / rest** slices so the shard planner can balance; the *rest* slice auto-covers future edges and asserts every identity check is claimed by exactly one slice. Structural fix (spectrum cache) filed as #707 |
| 2 | `check_duality_maps`: malformed `param_map` image fell through to the involution probe → possible uncaught throw inside `coherence_report` | `continue` after the type-mismatch finding; `involution=true` now **requires `source === target`** (validated in `dual!`, tested); second `param_map` application guarded with try/catch → broken edges report findings, never crash |
| 3 | SU(2) isotropy tested 3 ways (harness + per-model files + generated) | deletion criterion recorded in `src/identity_registry.jl` (derivative identities generatable + zero-value checks + Δ-conditional gates per #706); overlap is the #698 parity reference until then |
| 4 | canonical-row scan copy-pasted ×4 across duality/limits | kernel helper `_canonical_row` — one definition |
| 5 | `exclusions` encode registry-metadata gaps (kwargs-swallow ×3, 2D materialization ×2) at edge altitude | kept (loud-fail, never silent) — trait promotion filed as #705; the concrete SSH/TightBinding finding posted to #508 |
| 6 | Energy-granularity equivalence hard-coded in the kernel | `_row_covers` now routes through the `_equivalent_rows` **dispatch extension point** — new equivalence axes add a method next to the owning quantity, zero kernel edits |
| 7 | `@symmetry gapped` vs registered `MassGap` = two sources of truth, contradictions silent | new **`:symmetry` generated kind**: every declared `gapped` fact with an independent `MassGap@Infinite` row is corroborated by fetch (5 checks materialize — Heisenberg gapless ✓, S1 Haldane gap ✓, …) |
| 8 | latent duplicate-id crash for isotropy over families with shared components (parametric `XXCorrelation{M}` …) | isotropy keeps **one representative per component** — also removes the physically wrong same-component comparison |
| 9 | Δ-conditional symmetry enhancement (XXZ at Δ=1) inexpressible | filed as #706 (additive `at`-predicate fields, the @realizes pattern) |
| 10 | limit monotonicity magic numbers (1.1, 1e-14) | `mono_slack` is a declared edge field (default 0.1); absolute floor named `_MONO_ATOL_FLOOR` |

Local validation after fixes: **119 pass / 9 declared-skip / 0 fail**, coherence 0 findings. Generated totals: 70 identity / 52 dual / 1 limit / 5 symmetry.

sotashimozono:
## Second review pass — multi-agent (`/review-pr`) — fixed in d9f3aab24a

Six specialized agents (code-reviewer / comment-analyzer / pr-test-analyzer / silent-failure-hunter / type-design-analyzer / code-simplifier). One physics-comment finding was a **false positive** — `σˣ = 1 − 2n` is correct for this codebase's *x*-field TFIM convention (`H = -J σᶻσᶻ - h σˣ`), the textbook *z*-convention doesn't apply — verified and dropped. Everything else addressed:

**Important**
- **I-1** C11 coverage: an ungated component-isotropy identity with no implementing hub now self-reports a `:gap` (was silently uncovered) — coverage dispatches per edge type
- **I-2** `CheckOutcome` gains `:error`: a runner that *threw* (config/dispatch bug) is no longer mislabeled `:fail` (numerical disagreement)
- **I-3** synthetic store mutations wrapped in `try/finally` (a failing assertion can't leak a probe into a global store)
- **I-4** C10b symmetry corroboration now has a real *contradiction* test (flip Heisenberg1D → `gapped=true` vs its gap=0 ⇒ `:fail`), not just a pass assertion

**Advisory**
- **A-1** `Δ→1⁺` header comment → `Δ→1⁻` (matches the approach sequence)
- **A-2** docstrings: `dual!` involution `source===target`; `identity!` finite_N spin-1 trap; `_equivalent_rows` location guidance de-contradicted
- **A-3** tolerance guards: `rtol ∈ [0,1)` (rtol≥1 passes everything), `atol/final_atol ≥ 0` — shared `_check_tolerances`
- **A-4** `generated_checks` names the culprit kind on a throwing generator + the colliding ids; `references_of` default tolerates a field-less future store
- **A-5** C10b: gapped profile with no independent `MassGap@Infinite` ⇒ `:gap` (consistent with C12/C13), wired into `coherence_report`
- **A-6** `test_identity_rest` asserts each claimed prefix matches ≥1 check (was a tautology)

**Simplification** — `_both_endpoints_independent`, `_check_endpoint_rows!` (collapse the duplicated duality/limit scan + C12/C13 gap loops), `_point_suffix`, `_push_excluded_check!`, `_THERMAL_SWEEP`.

**Type design** — **T-1** `IdentityEdge` split into `TupleIdentityEdge`/`IsotropyIdentityEdge` under `AbstractIdentityEdge` (illegal mode/inactive-half states now unrepresentable; queries+generators dispatch). **T-2** folded into the `:error` status (a 4-way sum-type hierarchy would force dispatch in the reporter loop for no gain at this scale).

Local: coherence **0 errors / 0 gaps**; generated **119 pass / 9 skip / 0 fail / 0 error**; I-4 contradiction ⇒ `:fail`.

Issue #721, opened 2026-07-07
search: finalize query conventions (regime↔axes consistency), self-describing query schema, and enrich Hit (params/notes/valid_domain)


sotashimozono:
## Design review of the availability-search (`#711`/`#712`/`#713`), from an LLM's seat

I drove the search API (`feat/describe`, v0.25.4) as an LLM agent would, on a real
task: *"to validate finite-size effects, which spin-spin correlations of the
Heisenberg family are available, in which models, at which boundary conditions?"*
It **works, and works well for the primary path** — but the live run surfaced one
dangerous inconsistency and a set of enrichments that would make it a much
stronger LLM tool. Filing this to (1) pin down the query conventions, (2) make
those conventions self-describing, and (3) enrich `Hit`.

### What works (keep)
- `search(quantity=:correlation)` / `search(quantity=AbstractTwoPointCorrelation)` →
  13 Heisenberg / 32 atlas-wide correlation hubs. Exactly right.
- **BC is a first-class facet** and cleanly splits the finite-size question:
  `Heisenberg @ OBC` = 46 (correlations, local obs), `@ Infinite` = 38 (TDL
  quantities, structure factors), `@ PBC` = 3 (ConformalTower). This *is* the
  finite-size-vs-TDL axis.
- `describe_jsonl` is an excellent grounding payload (summary + LaTeX Hamiltonian +
  quantities + relations + refs); `realizing(:Heisenberg)` → HaldaneShastry /
  Heisenberg1D / XXZ1D; trust signals (`method`, `status`, `cross_checked`,
  `verified`) are all present in JSONL.

---

### Problem 1 — the `regime` facet silently drops whole quantity classes  ⟶ finalize the rules

Empirically (Heisenberg, live):

```
regime=ground_state        → 21 hits, correlation surfaced? false
regime=finite_temperature  → 16 hits, correlation surfaced? false
regime=universality        → 16 hits, correlation surfaced? false
regime=dynamics            →  0 hits            (velocity-only; TFIM's XXCorrelation{:dynamic} not surfaced)
```

`REGIMES` is a hand-picked predicate set —
`ground_state = Gap|Entanglement|GroundStateEnergyDensity`,
`finite_temperature = ThermalPotential`, `dynamics = Velocity` — so
**correlations, structure factors, magnetization, and susceptibility belong to no
regime at all**. An LLM that reaches for `regime=:ground_state` to get
[...]

PR #713, opened 2026-06-25
feat: describe(model) + realizing(class) — richer use-face read queries

sotashimozono:
## What (stacked on #712)

Two richer use-face read queries — the LLM's grounding + disambiguation surface.

### `describe(model)` — the full per-model grounding record
`describe(model) -> Vector{ModelRecord}` / `describe_jsonl([io], model)`. For each matching model:
`{summary, hamiltonian, quantities, relations, references}` — everything the atlas records, in one
record an LLM grounds a physics statement on. Composes the `@about` card + the search/relations layers.

**Disambiguation via structure** (the dogfood FibonacciAnyons trap): about-cards are sparse (23/80)
and FibonacciAnyons has none — but its record
`{summary:"", quantities:["TopologicalEntanglementEntropy"], references:[FreedmanKitaevLarsenWang2003, KitaevPreskill2006, ReadRezayi1999]}`
identifies it as a topological anyon model, **not** a Fibonacci quasicrystal. The structural content
disambiguates where prose is absent. (Universal one-line summaries = a separate `@about` coverage task.)

### `realizing(class)` — the class→models inverse
`realizing(class) -> Vector{Relation}` / `realizing_jsonl([io], class)`. The inverse of a model's
`:realizes` edge: which physical models realize a universality class — the dogfood scenario-4 question,
answered directly. `realizing(:Ising)` → TFIM / IsingSquare / IsingTriangular / Kitaev1D / ZnClock /
ZnParafermion, each with its critical locus + CFT data + references. Case-insensitive exact class match
(does not catch `IsingSDRG`).

**73 assertions** green on `main` (describe 10, realizing 8). `version 0.25.3 → 0.25.4`.

> Do not merge — stacked on #712, held for design review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #723, opened 2026-07-07
feat(query): enrich Hit with fetch call signature + notes/valid_domain (#721)

sotashimozono:
# feat(query): enrich `Hit` with the fetch call signature + notes/valid_domain

> **Stacked on #713** (`feat/describe`). Addresses part 1 (the cheap, high-value
> half) of the search design review in **#721**.

## Why

Driving the availability-search as an LLM agent (see #721) showed that a `Hit`
answers *"is it available?"* but not *"how do I obtain it?"* or *"where does it
hold?"* — even though that information is already recorded. To actually **measure**
a Heisenberg spin-spin correlation for finite-size validation, a consumer needs
the fetch's call signature (`i`, `j`, `beta`, `N`, …) and the closed form's
validity — none of which `Hit` surfaced.

## What

`Hit` (and the JSONL hit object) gain three additive fields:

- **`params`** — the fetch's declared keyword arguments, i.e. the **call
  signature**: "available" → "available *and here is how to call it*". Introspected
  via `Base.kwarg_decl` (the same mechanism `axes.jl` already uses for the thermal
  axis); the `kwargs...` slurp is dropped, result is sorted-unique. This also
  answers the **system-size** question: a fetch that declares `N`/`finite_N` is
  evaluable at any finite size (OBC/PBC hubs are finite-N closed forms; `Infinite`
  is the TDL) — otherwise invisible, as the atlas has no `N` registry dimension.
- **`notes`** / **`valid_domain`** — the `Implementation` row's remark and where
  the value holds (`""` when unset), so a consumer can match its numerics to the
  functional form / validity range.

Example (JSONL, `search_jsonl(; model=:TFIM, quantity=:SpecificHeat)`): each hit
now carries `"params":["beta",…]`, `"notes":"…"`, `"valid_domain":"…"` alongside
the existing provenance keys.

## Test

Extends `test/core/test_query.jl`: every hit has well-typed `params`/`notes`/
`valid_domain`; `params` is sorted-unique with no `kwargs` slurp; a
finite-temperature quantity exposes its temperature kwarg; JSONL surfaces the new
keys. Purely additive — existing keys/shape unchanged.

## Scope

This is item 1 of #721. The consistency fix (regularize `regime` ↔ axes so no
quantity class silently drops — the `regime=:ground_state` returns-no-correlations
trap) and the self-describing `query_schema()` are proposed in #721 for discussion
before implementing.

Version `0.25.4 → 0.25.5` (vs the `feat/describe` base).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "QAtlas"
 uuid = "2bd488d2-d3e6-46ee-afb3-5515643db0c7"
-version = "0.25.4"
+version = "0.25.5"
 authors = ["sota shimozono <shimozono-sota631@g.ecc.u-tokyo.ac.jp>"]
 
 [deps]
--- a/src/core/query.jl
+++ b/src/core/query.jl
@@ -34,10 +34,19 @@ const REGIMES = (
 """
     Hit
 
-One available (model, quantity, bc) hub, with its provenance: `method` (how the value is obtained),
-`status`, `reliability`, `references`, and `cross_checked` — `true` when the model participates in
-an executable cross-check (a duality, limit, or LSM-symmetry edge), the QAtlas-specific trust signal
-(model-level in this version).
+One available (model, quantity, bc) hub, with its provenance AND how to obtain the
+value. Provenance: `method`, `status`, `reliability`, `references`, `cross_checked`
+(`true` when the model is in an executable cross-check — a duality, limit, or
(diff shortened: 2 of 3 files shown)


sotashimozono:
Superseded: this Hit-enrichment content landed on main via the collapsed search merge #725 (main is now 0.25.1 with the full search feature). Closing as merged-through.

PR #758, opened 2026-07-21
feat: @response edges + fix AD through a non-differentiable root-finder

sotashimozono:
## Proposed Changes

Third constraint edge kind, after `@identity` (equalities) and `@bound` (inequalities). **`@response`** covers the AbstractQAtlas relations stated with a **supplied derivative** — `EntropyResponse(S, dF_dT)`, `SpecificHeatFromEntropy(C, dS_dT, T)` — which the other two cannot express because one slot is not fetchable. **132 checks over 28 hubs.**

A relation turns out to have **three** kinds of slot, and telling them apart is most of the work:

| slot | example | supplied from |
|---|---|---|
| quantity | `S::ThermalEntropy` | `fetch` |
| **field** | `T::Temperature` — `AbstractField`, *not* `AbstractQuantity` | the sweep point (it is the state, not an observable) |
| untyped | `dF_dT` | differentiation (#757's supplier) |

Subject-vs-derived, not residual-vs-zero, so both sides of a failure stay physical numbers — same shape as `:gibbs`.

## The part that matters

Wiring this up surfaced a real defect **and a false claim I had already merged**.

### 1. AD was silently wrong, and the model is now fixed

`CurieWeissIsing` solves its self-consistency by **bisection**. Bisection is the right *solver* and the wrong thing to *differentiate*: the β-dependence of its answer flows through the bracket endpoints and the comparison branches, not through `g(m) = 0`.

| T | fetched `C` | AD `T·dS/dT` (before) | FD `T·dS/dT` |
|---|---|---|---|
| 0.5 | 0.36594804 | **−0.01267792** | 0.36594804 |
| 1/3 | 0.09345921 | **−0.00438224** | 0.09345921 |

The **wrong sign**, on a value itself converged to 1e-14. Fixed properly, in the model: solve on primals (still bisection), then re-attach the derivative with **one Newton step from the converged root**. `g(m₀) = 0` there, so the value is untouched; the dual part becomes `−(∂g/∂β)/(∂g/∂m)` — the implicit-function theorem. The primal-stripping hook `_primal` comes from the **ForwardDiff extension**, so this needs no hard dependency: an unplanned second use of #757's ext machinery.

### 2. Correcting a claim already on main

#757 shipped *"Forward-mode AD is exact to round-off"* in `default_rtol`'s docstring. **That is false** for a fetch containing a root-find or an iterative solve. Corrected, with the measured table, and the correction says plainly that a tolerance cannot rescue a wrong derivative.

### Two defences for what the model fix cannot reach

- **`derivative_agreement`** cross-checks the derivative against an independent method. A disagreement is evidence about the **method**, so it is reported as a visible skip, never as a failed physical identity. It catches non-differentiable points for free — at a transition FD straddles the jump while AD takes a one-sided limit. At the Curie-Weiss transition `∂g/∂m → 0` and `dm/dβ` genuinely diverges; the fix lets that come out non-finite rather than fabricate a finite wrong number, and this catches it.
- **`rtol_floor`** — an identity cannot be verified more tightly than the values it relates. IsingSquare's Onsager `C` and `S` are quadrature evaluations and reproduce `C = T ∂S/∂T` to **2.5e-5, with AD and with finite differences alike, to the same figure** — so that is the accuracy of the *fetches*. AD's 1e-6 was reporting a quadrature residue as a physics failure.

## Usage or Results

Measured along the way: **XXZ1D's Klümper NLIE passes under AD to 8 digits** at both OBC and Infinite. So *"iterative solve ⇒ AD unsafe"* is too coarse — what matters is whether the method has a meaningful derivative at all. The simplest scalar root-find was the broken one; the heavy integral equation was fine.

`version` → **0.28.6**.

> Verification is CI's job here. A local run covers only the files I remember to select, and today it twice went green on a selection CI then failed.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "QAtlas"
 uuid = "2bd488d2-d3e6-46ee-afb3-5515643db0c7"
-version = "0.28.5"
+version = "0.28.6"
 authors = ["sota shimozono <shimozono-sota631@g.ecc.u-tokyo.ac.jp>"]
 
 [deps]
--- a/ext/QAtlasForwardDiffExt.jl
+++ b/ext/QAtlasForwardDiffExt.jl
@@ -16,4 +16,8 @@ using ForwardDiff: ForwardDiff
 QAtlas.derivative(::ForwardDiffBackend, f, x::Real) = ForwardDiff.derivative(f, float(x))
 QAtlas.backend_available(::ForwardDiffBackend) = true
 
+# Lets a root-finder solve on primals and re-attach the exact implicit
+# derivative afterwards — see QAtlas._primal.
+QAtlas._primal(x::ForwardDiff.Dual) = ForwardDiff.value(x)
+
 end
--- a/src/QAtlas.jl
+++ b/src/QAtlas.jl
@@ -60,6 +60,9 @@ using AbstractQAtlas:
     # ...and the inequality verbs, so a BOUND edge (core/bound.jl) can delegate
     # its criterion to AbstractQAtlas exactly as :gibbs delegates its arithmetic.
     AbstractInequality,
+    AbstractRelation,
+    EntropyResponse,
+    SpecificHeatFromEntropy,
     slack,
     variable_slots,
     # the inequalities bound_registry.jl declares edges for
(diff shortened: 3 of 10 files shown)


sotashimozono:
## Verification of the implicit-differentiation fix

Measured after the change (the PR body carries only the "before" table):

| T | fetched `C` | AD `T·dS/dT` **after** | FD `T·dS/dT` | AD **before** |
|---|---|---|---|---|
| 2.0 | 0.0 | 0.0 | 0.0 | 0.0 |
| 0.5 | 0.365948044 | **0.365948044** | 0.36594804 | −0.01267792 |
| 1/3 | 0.0934592109 | **0.0934592109** | 0.0934592137 | −0.00438224 |

Nine digits, and at `T = 1/3` **more accurate than the finite difference** (0.0934592109 vs 0.0934592137) — expected, since the implicit derivative carries no truncation error. The wrong-sign result is gone.

The full-suite verdict is CI's.

PR #760, opened 2026-07-21
feat: model-axis derivatives + opt-in model lists (MagnetizationResponse, SusceptibilityResponse)

sotashimozono:
## Proposed Changes

Derived inputs can now differentiate along a **model field**, not just the state point: `wrt` is `:T`/`:β` (vary the fetch kwargs) or any model field (rebuild the model). The rebuild is **`_with_param`, which already existed** for `@limits_to`'s `param = :Δ` walk — so this reuses the atlas's own parameter mechanism rather than inventing a second one.

Unlocks `MagnetizationResponse` (`M = −∂F/∂h`) and `SusceptibilityResponse` (`χ = ∂M/∂h`).

### 1. A model axis is finite-difference only

Model structs type their parameters concretely:

```julia
struct TFIM <: AbstractQAtlasModel
    J::Float64
    h::Float64
end
```

so rebuilding one with an AD dual runs it through `Float64(...)` and **drops the derivative silently**. The axis therefore carries its own backend and its own tolerance, and the edge is reported at the loosest of them.

Making AD work here means parameterizing the model structs — a breaking change to a type dispatched on in ~50 places. That belongs in its own PR, not smuggled in.

### 2. These relations are not universal, so `models` is an ALLOW-list

AbstractQAtlas types the subject `Magnetization{:z}`, so `M = −∂F/∂h` holds only where `h` is the **longitudinal** field:

| coupling | models | valid? |
|---|---|---|
| `−h Σ σᶻ` | `IsingChain1D`, `CurieWeissIsing`, `LongRangeXY1D` | ✅ |
| `−h Σ σˣ` | `TFIM`, `LongRangeIsing1D` | ❌ — `−∂F/∂h` is `⟨σˣ⟩`, not `M_z` |

**`derivative_agreement` cannot catch this.** Both backends agree on the wrong number; the cross-check detects a bad *method*, never a relation applied to a Hamiltonian it does not describe. So the defence has to be structural.

Opt-**in** rather than exclusion means a model added later is silently *skipped* instead of silently *checked against physics that does not apply to it* — the same "do nothing rather than be wrong" default the rest of this layer uses. There is a test asserting `TFIM` is absent from these checks and present in the universal ones.

## Usage or Results

```julia
∂(FreeEnergy, :h; then = d -> -d)   # M = -∂F/∂h, model axis ⇒ finite differences
∂(Magnetization{:z}, :h)           # χ = ∂M/∂h
```

`version` → **0.28.8**.

> Stacked on #759 (base `main`). Verification is CI's.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "QAtlas"
 uuid = "2bd488d2-d3e6-46ee-afb3-5515643db0c7"
-version = "0.28.7"
+version = "0.28.8"
 authors = ["sota shimozono <shimozono-sota631@g.ecc.u-tokyo.ac.jp>"]
 
 [deps]
--- a/src/QAtlas.jl
+++ b/src/QAtlas.jl
@@ -65,6 +65,8 @@ using AbstractQAtlas:
     SpecificHeatFromEntropy,
     GibbsHelmholtz,
     SpecificHeatFDT,
+    MagnetizationResponse,
+    SusceptibilityResponse,
     slack,
     variable_slots,
     # the inequalities bound_registry.jl declares edges for
--- a/src/core/response.jl
+++ b/src/core/response.jl
@@ -44,12 +44,24 @@ than a symbolic mini-language: the whole point of this layer is that the algebra
 lives in AbstractQAtlas, so anything expressible here should stay small enough to
 read at the declaration site.
 
-Only the temperature axis is supported here.  A field derivative (`dF/dh`, for
-`MagnetizationResponse`) is deliberately absent: `h` is a MODEL parameter, not a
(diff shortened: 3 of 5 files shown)


sotashimozono:
Superseded by #778, which reworks this onto current main and fixes three things that only showed up by running the checks it generates: a stale `backend` reference that made 48 `entropy_response` checks error, a double sign flip (`MagnetizationResponse` is `M = -dF_dh`, so the extra `then = d -> -d` asserted `M = +∂F/∂h` — invisible at h = 0 where both sides vanish), and the fact that the default model is the useless place to evaluate a field response. #778 adds `at = (h=0.35,)` for that last one.

